PR #1, opened 2016-04-11
Upstream updates bugfixes

timvaillancourt:
Fixes:

1. Index counter bugfix from upstream.
2. Improved version of my direct-connection fix (for replsets) from upstream.
3. Updated the broken tests of upstream code, which was updated in upstream.
4. mongos/replset/mongod logic in mongodb_collector.go moved to functions.
5. Lots of code cleanup, added logging, etc.


collector/mongod/index_counters.go
```diff
@@ -23,7 +23,7 @@ var (
 
 //IndexCounterStats index counter stats
 type IndexCounterStats struct {
-	Accesses  float64 `bson:"accesses`
+	Accesses  float64 `bson:"accesses"`
```
dbmurphy: How did this ever work?

collector/mongod/metrics.go
```diff
@@ -397,7 +397,7 @@ type CursorStats struct {
 // Export exports the cursor stats.
 func (cursorStats *CursorStats) Export(ch chan<- prometheus.Metric) {
 	metricsCursorTimedOutTotal.Set(cursorStats.TimedOut)
-	metricsCursorOpen.WithLabelValues("noTimeout").Set(cursorStats.Open.NoTimeout)
+	metricsCursorOpen.WithLabelValues("timed_out").Set(cursorStats.Open.NoTimeout)
```
dbmurphy: Are we sure noTimeout means timed_out,  sounds like that means its not timed out to me

collector/mongodb_collector.go
```diff
@@ -40,146 +36,80 @@ func NewMongodbCollector(opts MongodbCollectorOpts) *MongodbCollector {
...
-        }
-	session, err := mgo.Dial(uri)
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "error: %v\n", err)
-        os.Exit(1)
-		glog.Errorf("Cannot connect to server using url: %s", uri)
-		return nil,err
-	}
-
-	session.SetMode(mgo.Eventual, true)
```
dbmurphy: Why did we drop eventual and the disabling of timeouts?

collector/shared/version.go
```diff
@@ -1,42 +0,0 @@
-package collector_shared
```
dbmurphy: I dont know about this is looks like  version checking functions are broken which will be needed to say if 3.0 do X , if 3.2 do y , if 3.4  do z right

mongodb_exporter.go
```diff
@@ -3,21 +3,30 @@ package main
...
+	return "mongodb://localhost:27017"
+}
+
 var (
 	listenAddressFlag = flag.String("web.listen-address", ":9001", "Address on which to expose metrics and web interface.")
 	metricsPathFlag   = flag.String("web.metrics-path", "/metrics", "Path under which to expose metrics.")
 
-	mongodbURIFlag    = flag.String("mongodb.uri", "mongodb://localhost:27017", "Mongodb URI, format: [mongodb://][user:pass@]host1[:port1][,host2[:port2],...][/database][?options]")
+	mongodbURIFlag    = flag.String("mongodb.uri", mongodbDefaultUri(), "Mongodb URI, format: [mongodb://][user:pass@]host1[:port1][,host2[:port2],...][/database][?options]")
 	enabledGroupsFlag = flag.String("groups.enabled", "asserts,durability,background_flushing,connections,extra_info,global_lock,index_counters,network,op_counters,op_counters_repl,memory,locks,metrics", "Comma-separated list of groups to use, for more info see: docs.mongodb.org/manual/reference/command/serverStatus/")
```
dbmurphy: Are we still actually using  group flags?  I would kind of like to do away with that concept  and  have --filter-checks "index_counters,op_counters,..."  to be inclusive by default and filter first on discovered type then on filters provide from config

collector/mongod/index_counters.go
timvaillancourt: Not entirely sure of the inner workings here but I think it would not match the bson output and therefore always report accesses as zero, but I am not 100% sure. Structs seem very forgiving of missing/undefined fields

collector/mongodb_collector.go
timvaillancourt: Both of those were moved to the 'MongoSession' func in [shared/connection.go](https://github.com/timvaillancourt/prometheus_mongodb_exporter/blob/upstream_updates_bugfixes/shared/connection.go#L15-34)

collector/mongod/metrics.go
timvaillancourt: Very interesting, no we are not :)

"metrics.cursor.open.noTimeout
New in version 2.6.

The number of open cursors with the option DBQuery.Option.noTimeout set to prevent timeout after a period of inactivity."

In this case I think the original "noTimeout" is more appropriate, to represent the noTimeout driver option.

collector/shared/version.go
timvaillancourt: Currently there is no code that used those functions, so I removed them. The code currently doesn't need to behave differently (pending any bugs) between versions due to the struct based output, and the "if == nil" checking on each subcomponent before it is Exported to Prometheus. For eg: the WiredTiger code works perfectly on MongoDB 2.6 because the code notices it is not in the db.serverStatus() output, and therefore never exports it. I'm hoping we can continue to use this approach unless there is a blocker.

mongodb_exporter.go
timvaillancourt: I've never touched the group flags, but they're there - still untouched mainline code.

collector/mongod/metrics.go
timvaillancourt: I've reverted this back to 'noTimeout' to match the driver setting. To conclude: this metric is telling us the current number of connections who are using the "DBQuery.Option.noTimeout" option.

mongodb_exporter.go
dbmurphy: Yeah but we are not updating the groups file for all the new groups we added, might be "stale" at this point

collector/shared/version.go
dbmurphy: I would like to have a version compare functionality but no nothing "has" to have it now, but are losing a feature we may need later.

timvaillancourt: That is a good point, we may need it later. Let me add this back somewhere for future-proofing.

timvaillancourt: It turns out I already preserved the the version checking function in [shared/connection.go](https://github.com/timvaillancourt/prometheus_mongodb_exporter/blob/upstream_updates_bugfixes/shared/connection.go#L36-43), so the only thing left was the comparison function. I've added it to shared/utils.go (a better place for it anyways) in this commit:

https://github.com/timvaillancourt/prometheus_mongodb_exporter/commit/37ec32bf5e157010df5c3493fbb79762752074ea

mongodb_exporter.go
timvaillancourt: Yeah it could be stale/few-use-cases. I don't think it's harming anything by being there but I'm ok with removing it also.

PR #3, opened 2016-04-13
Tims code updates

dbmurphy:


collector/mongod/oplog_status.go
```diff
@@ -1,102 +1,116 @@
 package collector_mongod
 
 import (
-    "time"
```
dbmurphy: Do we not have a need for time any more?

timvaillancourt: No, we get a compile error because it is imported and not used

collector/mongod/oplog_status.go
```diff
@@ -1,102 +1,116 @@
...
+	})
+	oplogStatusSizeBytes = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace:	Namespace,
+		Subsystem:	"replset_oplog",
+		Name:		"size_bytes",
+		Help:		"Size of oplog in bytes",
+	}, []string{"type"})
 )
 
-func GetCollectionSizeMB(db string, collection string, session *mgo.Session) (float64) {
```
dbmurphy: Why are we removing this? Displaying it in bits could make it an int64 vs int32 right?

collector/mongod/oplog_status.go
```diff
@@ -1,102 +1,116 @@
...
-    }
-
-    minTime := ParseBsonMongoTsToUnix(results.MinTime)
-    maxTime := ParseBsonMongoTsToUnix(results.MaxTime)
-
-    now := time.Now().Unix()
-    lengthSeconds := maxTime - minTime
-    lengthSecondsNow := now - minTime
-
-    return float64(lengthSeconds), float64(lengthSecondsNow)
```
dbmurphy: We are losing some scope here we no longer know now - first time stamp. As discussed before this is causing use to store more data than is needed and more than competing mongo monitoring software stores.

collector/mongod/oplog_status.go
timvaillancourt: The only lowlevel datatype in prometheus is float64, everything has to be one even if it's a 0/1. It won't accept regular float/int or int64.

collector/mongod/replset_status.go
```diff
@@ -1,232 +1,223 @@
 package collector_mongod
 
 import (
-    "time"
-    "github.com/golang/glog"
-    "github.com/prometheus/client_golang/prometheus"
-    "gopkg.in/mgo.v2"
-    "gopkg.in/mgo.v2/bson"
+	"time"
+
```
dbmurphy: Why the empty line returns?

collector/mongod/replset_status.go
```diff
@@ -1,232 +1,223 @@
...
-            Name:      "max_n2n_ping_ms",
-            Help:      "Maximum ping in milliseconds to other replica set members",
-    })
+	myState   = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: Namespace,
+		Subsystem: "replset",
+		Name:      "my_state",
+		Help:      "An integer between 0 and 10 that represents the replica state of the current member",
+	}, []string{"set"})
+	term = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
dbmurphy: Should we not name this to be more human friendly unless someone know mongo they dont know what term means. Last election_id might be a better name for graphs

collector/mongod/replset_status.go
```diff
@@ -1,232 +1,223 @@
...
+		Subsystem: "replset",
+		Name:      "heatbeat_interval_millis",
+		Help:      "The frequency in milliseconds of the heartbeats",
+	}, []string{"set"})
+	memberHealth = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: Namespace,
+		Subsystem: "replset",
+		Name:      "member_health",
+		Help:      "This field conveys if the member is up (1) or down (0).",
+	}, []string{"set", "name", "state"})
```
dbmurphy: Again we have scope creep we are add more metrics. we should only monitor what is required, and have meetings on anything  additional

collector/mongod/wiredtiger.go
```diff
@@ -0,0 +1,403 @@
...
+		stats.Cache.Export(ch)
+	}
+	if stats.Transaction != nil {
+		stats.Transaction.Export(ch)
+	}
+	if stats.ConcurrentTransactions != nil {
+		stats.ConcurrentTransactions.Export(ch)
+	}
+
+	wtBlockManagerBlocksTotal.Collect(ch)
```
dbmurphy: Does MMS collect all of this? Are we collecting to much? Not enought?

collector/mongos/sharding_status.go
```diff
@@ -0,0 +1,245 @@
...
+			}
+		}
+	}
+	balancerIsEnabled.Set(status.IsBalanced)
+	balancerChunksBalanced.Set(status.BalancerEnabled)
+
+	balancerIsEnabled.Collect(ch)
+	balancerChunksBalanced.Collect(ch)
+	mongosClusterIdHex.Collect(ch)
+	mongosClusterCurrentVersion.Collect(ch)
```
dbmurphy: Seems over kill to watch a version with will only change on an upgrade of major release

collector/mongos/sharding_status.go
```diff
@@ -0,0 +1,245 @@
...
+	mongosClusterMinCompatVersion.Collect(ch)
+	mongosUpSecs.Collect(ch)
+	mongosPing.Collect(ch)
+	mongosBalancerLockState.Collect(ch)
+	mongosBalancerLockTimestamp.Collect(ch)
```
dbmurphy: Not sure this really matters

collector/mongod/oplog_status.go
timvaillancourt: The previous code had 2 x values for oplog time: "length" and "length to now", which are 2 x float64 values. This new code now returns "head" and "tail" which is also 2 x float64 values that can provide more flexibility in graphing in the future. I feel this is a more optimal approach but this can be rolled back if we must, however this will require updates and retesting of the grafana_mongodb_dashboards project also.

collector/mongos/sharding_status.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+	}
+	balancerIsEnabled.Set(status.IsBalanced)
+	balancerChunksBalanced.Set(status.BalancerEnabled)
+
+	balancerIsEnabled.Collect(ch)
+	balancerChunksBalanced.Collect(ch)
+	mongosClusterIdHex.Collect(ch)
+	mongosClusterCurrentVersion.Collect(ch)
+	mongosClusterMinCompatVersion.Collect(ch)
```
dbmurphy: Seems over kill

collector/mongos/sharding_status.go
```diff
@@ -0,0 +1,245 @@
...
+				mongosBalancerLockState.With(labels).Set(status.BalancerLock.State)
+			}
+		}
+	}
+	balancerIsEnabled.Set(status.IsBalanced)
+	balancerChunksBalanced.Set(status.BalancerEnabled)
+
+	balancerIsEnabled.Collect(ch)
+	balancerChunksBalanced.Collect(ch)
+	mongosClusterIdHex.Collect(ch)
```
dbmurphy: Why do we need the cluster Id graphed it wont change

collector/mongod/replset_status.go
timvaillancourt: replset_status.go is an improved version of my code pulled down from mainline, so this was a choice from the upstream committer, unfortunately. It can be removed

collector/mongod/replset_status.go
timvaillancourt: The area I see highlighted here is "myState" which seems to be a good name to me. This is upstream code so I tried not to mess with it, but I am welcome to change this to anything if it doesn't cause incompatibility with the graph templates

collector/mongos/sharding_topology.go
```diff
@@ -0,0 +1,181 @@
...
+)
+
+var (
+	shardingTopoInfoTotalShards = prometheus.NewGauge(prometheus.GaugeOpts{
+			Namespace:	Namespace,
+			Subsystem:	"sharding",
+			Name:		"shards_total",
+			Help:		"Total # of Shards in the Cluster",
+	})
+	shardingTopoInfoDrainingShards = prometheus.NewGauge(prometheus.GaugeOpts{
```
dbmurphy: This is over kill also

collector/mongod/replset_status.go
timvaillancourt: replset_status.go is an improved version of my code pulled down from mainline, so this was a choice from the upstream committer. There is no extra work being done to gather these metrics seeing they're already in the result we require, but we can disable anything you think is extraneous if they're not used in graphs and space is a concern

collector/mongod/wiredtiger.go
timvaillancourt: I'm not sure on all accounts. I started small and added until I felt like I could "see enough" of what WT was doing. I would say this is the bare minimum we might need, and more can be added.

collector/mongos/sharding_status.go
timvaillancourt: Removing quite a few of those would be breaking changes to the grafana templates..which ones do you mean, or all?

collector/mongos/sharding_status.go
timvaillancourt: Removing quite a few of those would be breaking changes to the grafana templates..which ones do you mean, or all?

collector/mongos/sharding_status.go
timvaillancourt: It's gathered but not graphed. This can be removed if there is concern

collector/mongos/sharding_topology.go
timvaillancourt: The # of shards is used in several graphs, and I feel it is useful, no?

collector/mongos/sharding_status.go
timvaillancourt: Edit: ok, I'll disable the 'Version' section of this code.

collector/mongod/replset_status.go
dbmurphy: I was actually highlighting term = prometheus.NewGaugeVec(prometheus.GaugeOpts{ but differing with due to the beta tag

collector/mongos/sharding_topology.go
dbmurphy: No I was talking about  +			Name:		"shards_draining_total", (I had highlighted the  line that was the start of the function, so was not talking about shards_total. Remoing comment we will fix this later as you added the beta tag

PR #4, opened 2016-04-13
Removing all WT Async code because the metrics do nothing (always …

timvaillancourt:
…zero)

Also removed traces of 'Cursor' and 'Connection' stats for WT that are disabled but still present.


timvaillancourt:
Consolidating these wiredtiger improvements to one branch. Closing

PR #8, opened 2016-05-03
Change port from 9001 to 9104 by default

dbmurphy:
This is so the mysql and mongo exporters run on the same port by default. It can be changed in the configuration but  this simplifies things in other systems.


timvaillancourt:
Looks good to me!

PR #14, opened 2016-05-24
Revert "New Makefile from Prom upstream"

dbmurphy:
Reverts Percona-Lab/prometheus_mongodb_exporter#13

@roman-vynar 
This is rather broken on 2 different systems. We should raise an issue and track it, looking at the old common file it does almost all of what this does, but promu  is complaining about things that it shouldn't.Reverting to unblock @timvaillancourt


roman-vynar:
To build it with `promu` and as a best practice you have to place your code into `$GOPATH/src/github.com/Percona-Lab/prometheus_mongodb_exporter`.
Than it should build fine by doing `make build`, not `make` as it was before.
Up to you guys.

PR #17, opened 2016-07-17
clear the redundant information

wuxinwei:
clear the redundant information,avoid to influence alert in prometheus's metric

### DIFF
--- a/collector/mongod/replset_status.go
+++ b/collector/mongod/replset_status.go
@@ -141,7 +141,21 @@ func (replStatus *ReplSetStatus) Export(ch chan<- prometheus.Metric) {
 	if replStatus.HeartbeatIntervalMillis != nil {
 		heartbeatIntervalMillis.WithLabelValues(replStatus.Set).Set(*replStatus.HeartbeatIntervalMillis)
 	}
-
+	
+	// clear redundant information
+	myState.Reset()
+	term.Reset()
+	numberOfMembers.Reset()
+	heartbeatIntervalMillis.Reset()
+	memberState.Reset()
+	memberHealth.Reset()
+	memberUptime.Reset()
+	memberOptimeDate.Reset()
+	memberElectionDate.Reset()
+	memberLastHeartbeat.Reset()
+	memberLastHeartbeatRecv.Reset()
+	memberPingMs.Reset()
+	memberConfigVersion.Reset()
 	for _, member := range replStatus.Members {
 		ls := prometheus.Labels{
 			"set":   replStatus.Set,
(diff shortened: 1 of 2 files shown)


dbmurphy:
@timvaillancourt : Are these really needed given  all the metrics source data is gotten again, and an error will kill the request with an error. I am not sure resets are even needed. Same thing applies  to the connection as my thought was it will just fail on this  attempt but the next attempt will work fine.  Is there any risk here by not applying?

wuxinwei:
if I do not to reset those thing, It will return data contain the expired data, and mixed with new data,for example, once one mongod instance down, mongodb-exporter will send the info that mongod instance down info to prometheus monitor, if I seted alert rule, those info will trigger the alert, after mongod instance relived, the previous "mongod down" info still was deliveryed to prometheus monitor，and put the alert trigger, in fact, mongodb is normal, one genernate "mongod down " info , the alert will push alert info to my webhook unlimited, not only the mongodb_repset_member_health, but all those metric, I can reappear this situation

wuxinwei:
- first,I start a **prometheus_mongodb_exporter** and 4 mongods instance, and metric like this(see below)
```shell
./prometheus_mongodb_exporter -mongodb.uri "mongodb://localhost:27017" -web.listen-address ":37017"
### Warning: the exporter is in beta/experimental state and field names are very
### likely to change in the future and features may change or get removed!
### See: https://github.com/Percona-Lab/prometheus_mongodb_exporter for updates
mongodb_exporter version: unknown, git commit hash: unknown
Listening on :37017
```
[https://goo.gl/photos/b6DwV6H7b6mSHL5n6](url)

- then I stop one mongod instance,and metric shows one instance down
[https://goo.gl/photos/wKySadWBPgiNyuseA](url)

- and I stop the other one mongod instance, metric shows this info
[https://goo.gl/photos/74Gz9e6BNjekeqbNA](url)

- after I restart mongod instance, and those instances become health, the metric shows this:
[https://goo.gl/photos/9ktVKXBgxe5ULK9e9](url)

> I test those in my local machine, and metric just mixed the expired data and new data,so please check, thank you , @dbmurphy

wuxinwei:
Just Check my first pull request, please ignore the second

timvaillancourt:
@dbmurphy and @wuxinwei: I think I see what is being done here.

Yes, I have seen the Prometheus library return the "same result as last time" in some cases, this is probably what @wuxinwei is referring to.

While we could add .Resets() to metrics, I think we need to do this "all or nothing", not just in the replication sections - all areas of the upstream exporter and this exporter should be consistent.

dbmurphy:
Leaving this open until we have the new version of this merged in.

PR #31, opened 2016-11-14
Fixes to improve replication lag accuracy

timvaillancourt:
1. Reset metric vectors before setting them to clear stale replset states (ie: only display current replication state). PR #17 partially duplicates this change.
2. Add new metric: 'mongodb_mongod_replset_my_name' to show the 'name' of the replica set member. This will be used to join on the member_optime_date metric to select a more accurate metric for lag.

Example output post-change:
```
$ curl -s http://localhost:9104/metrics|grep replset_my_name
# HELP mongodb_mongod_replset_my_name The replica state name of the current member
# TYPE mongodb_mongod_replset_my_name gauge
mongodb_mongod_replset_my_name{name="localhost:27017",set="test1"} 1
```

### DIFF
--- a/collector/mongod/replset_status.go
+++ b/collector/mongod/replset_status.go
@@ -17,7 +17,7 @@ var (
 		Subsystem: subsystem,
 		Name:      "my_name",
 		Help:      "The replica state name of the current member",
-	}, []string{"name","set"})
+	}, []string{"name", "set"})
 	myState   = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: Namespace,
 		Subsystem: subsystem,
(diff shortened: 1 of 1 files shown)


timvaillancourt:
Hey @roman-vynar, I was just thinking. Instead of adding this 'self' vector in I could just add 1 x new metric like this:

mongodb_mongod_replset_my_name{name="host:port"}=1

And then we could just join on that instead of adding a new vector to several things.

Is that better or worse?

roman-vynar:
I think it's better than `self` label and can be used in other places as well if needed.

timvaillancourt:
Thanks @roman-vynar, let's do that instead - not sure why I didn't do this first!

Please put this merge on hold.

timvaillancourt:
Ok, all fixed. Please confirm this will work @roman-vynar:

```
$ curl -s http://localhost:9104/metrics|grep my_name
# HELP mongodb_mongod_replset_my_name The replica state name of the current member
# TYPE mongodb_mongod_replset_my_name gauge
mongodb_mongod_replset_my_name{name="localhost:27017"} 1
```

roman-vynar:
Looks fine 👍

PR #37, opened 2016-12-06
(fix) typo in the README MONGO_URI should be MONGO_URL

phil-taylor:
The README tells you to set MONGO_URI and this doesn't work. The code is looking for MONGO_URL, updated README to remove confusion.


timvaillancourt:
You're correct @phil-taylor. It is MONGO_URI, not MONGO_URL as seen here:

https://github.com/percona/mongodb_exporter/blob/master/mongodb_exporter.go#L108

I will merge this as it is README.md only.

PR #40, opened 2017-02-05
Enable tls connection with mongo server

### LINKED ISSUE #18: How to we configure it with ssl ?
I just wonder how we could configure it with ssl options ?

Also isn't this exporting compatible with rocksdb :cold_sweat: ?

ese:
Based on https://github.com/dcu/mongodb_exporter/pull/18
Fix #18

### DIFF
--- a/shared/connection.go
+++ b/shared/connection.go
@@ -3,6 +3,11 @@ package shared
 import (
 	"time"
 
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"net"
+
 	"github.com/golang/glog"
 	"gopkg.in/mgo.v2"
 )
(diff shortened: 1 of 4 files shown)


timvaillancourt:
I was in the process of merging this same PR, it works great!

I will run this through some testing before approving.

timvaillancourt:
Testing complete, no problems noticed. We will merge soon. Thanks

shared/connection.go
```diff
@@ -48,6 +68,74 @@ func MongoSession(uri string) *mgo.Session {
...
+		}
+		dialInfo.DialServer = func(addr *mgo.ServerAddr) (net.Conn, error) {
+			conn, err := tls.Dial("tcp", addr.String(), config)
+			if err != nil {
+				glog.Infof("Could not connect to %v. Got: %v", addr, err)
+				return nil, err
+			}
+			if config.InsecureSkipVerify {
+				err = enrichWithOwnChecks(conn, config)
+				if err != nil {
```
arvenil: Is this correct? Shouldn't this `return nil, err`? `enrichWithOwnChecks` returns closed connection on error. Returning error and closed connection might be misleading? This is not affecting anything so 👍

shared/connection.go
```diff
@@ -37,9 +51,15 @@ func MongoSession(uri string) *mgo.Session {
 	dialInfo.Direct = true // Force direct connection
 	dialInfo.Timeout = dialMongodbTimeout
 
+	err = opts.configureDialInfoIfRequired(dialInfo)
```
arvenil: I'm not a big fan of modifying values pass by reference. It's trickier to follow. Alternative:
```
err, dialServer := opts.dialServerFunc()
if err != nil {
	glog.Errorf("%s", err)
	return nil
}
dialInfo.DialServer = dialServer
```
It works so 👍

PR #45, opened 2017-02-21
Return 'mmapv1' as storage engine if db.serverStatus().storageEngine doesn't exist (added in 3.0)

timvaillancourt:
Our 'mongodb_mongod_storage_engine' metric seems to break on MongoDB 2.x because the db.serverStatus().storageEngine document did not exist in this version.

This changes the exporter to return 'mmapv1' as the storage engine type if db.serverStatus().storageEngine (3.0+ only) does not exist and/but db.serverStatus().backgroundFlushing does exist (only shows on MMAPv1 instances). The 2nd condition is to avoid TokuMX showing as MMAPv1, mostly.

EDIT: My golang IDE fixed some whitespace alignment on 2 x structs, no logic changed to those lines.

### DIFF
--- a/collector/mongod/server_status.go
+++ b/collector/mongod/server_status.go
@@ -128,7 +128,8 @@ func (status *ServerStatus) Export(ch chan<- prometheus.Metric) {
 		status.WiredTiger.Export(ch)
 	}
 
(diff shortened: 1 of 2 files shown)


percona-csalguero:
:+1:

PR #53, opened 2017-03-03
Use Primary readpreference (mgo.Strong) when using mongos to prevent SERVER-27864

timvaillancourt:
The only change is line 67, the rest is whitespace/lint fixes from my go IDE.

This fixes a problem that only happens on 3.2.10-3.2.12 with SCCC config servers by setting mongos connections to use Primary-only read preference (mgo.Strong).

MongoDB issue: https://jira.mongodb.org/browse/SERVER-27864

### DIFF
--- a/collector/mongodb_collector.go
+++ b/collector/mongodb_collector.go
@@ -1,10 +1,10 @@
 package collector
 
 import (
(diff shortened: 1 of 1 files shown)


percona-csalguero:
:+1:

PR #63, opened 2017-05-24
Build against latest version of prometheus golang-client

ese:
Resolved conflicts of #38

Issue #57, opened 2017-03-15
Cannot build


godjil:
# github.com/percona/mongodb_exporter/collector/mongos
/usr/local/go/src/github.com/percona/mongodb_exporter/collector/mongos/asserts.go:26: assertsTotal.WithLabelValues("regular").Set undefined (type prometheus.Counter has no field or method Set)
/usr/local/go/src/github.com/percona/mongodb_exporter/collector/mongos/asserts.go:27: assertsTotal.WithLabelValues("warning").Set undefined (type prometheus.Counter has no field or method Set)
/usr/local/go/src/github.com/percona/mongodb_exporter/collector/mongos/asserts.go:28: assertsTotal.WithLabelValues("msg").Set undefined (type prometheus.Counter has no field or method Set)
/usr/local/go/src/github.com/percona/mongodb_exporter/collector/mongos/asserts.go:29: assertsTotal.WithLabelValues("user").Set undefined (type prometheus.Counter has no field or method Set)
/usr/local/go/src/github.com/percona/mongodb_exporter/collector/mongos/asserts.go:30: assertsTotal.WithLabelValues("rollovers").Set undefined (type prometheus.Counter has no field or method Set)
/usr/local/go/src/github.com/percona/mongodb_exporter/collector/mongos/connections.go:36: connectionsMetricsCreatedTotal.Set undefined (type prometheus.Counter has no field or method Set)
/usr/local/go/src/github.com/percona/mongodb_exporter/collector/mongos/metrics.go:60: metricsGetLastErrorWtimeTotalMilliseconds.Set undefined (type prometheus.Counter has no field or method Set)
/usr/local/go/src/github.com/percona/mongodb_exporter/collector/mongos/metrics.go:62: metricsGetLastErrorWtimeoutsTotal.Set undefined (type prometheus.Counter has no field or method Set)
/usr/local/go/src/github.com/percona/mongodb_exporter/collector/mongos/metrics.go:80: metricsCursorTimedOutTotal.Set undefined (type prometheus.Counter has no field or method Set)
/usr/local/go/src/github.com/percona/mongodb_exporter/collector/mongos/network.go:32: networkBytesTotal.WithLabelValues("in_bytes").Set undefined (type prometheus.Counter has no field or method Set)
/usr/local/go/src/github.com/percona/mongodb_exporter/collector/mongos/network.go:32: too many errors
# github.com/percona/mongodb_exporter/collector/mongod
[...]

PR #38, opened 2016-12-12
Build against latest version of prometheus golang-client

jlemaes:
This pull request works again with the latest version of prometheus golang-client where in [this commit](https://github.com/prometheus/client_golang/commit/2fee50beaa152342e8347c41edf2202c3bd5f858) the counter.Set method was removed.

PR #67, opened 2017-06-15
Vendor dependencies

AlekSi:
This way exporter can be installed with `go get`.

Refs #57, #38, #63.


.travis.yml
```diff
@@ -0,0 +1,10 @@
+sudo: false
+language: go
+
+go:
+  - 1.7.x
+  - 1.8.x
+  - tip
+
+script:
+  - make build
```
arvenil: It doesn't have tests to run?

AlekSi: Tests are not even compiling in this branch 🤦‍♂️. They are already fixed in `develop` branch. The idea of that PR is to make exporter working and `go get`-able ASAP.

arvenil: ... so without tests... welcome to the dark side 🕶

PR #69, opened 2017-06-16
Release 0.2.0

AlekSi:
https://github.com/percona/mongodb_exporter/blob/release/0.2/CHANGELOG.md


CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
+# Changelog
+
+## v0.2.0 (2017-06-16)
+
+* Default listen port changed to 9216.
+* All log messages now go to stderr. Logging flags changed.
+* Fewer messages on default INFO logging level.
```
arvenil: I'm not sure about this one. If you disable logs for requests (because for every scrap request there was a log message) then you won't have information that exporter is processing them? Unless exporter is also instrumented and it delivers info about processed requests? I would like to avoid situation where exporter brakes but we won't know because we don't have any insights.

CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
...
+* Use https://github.com/prometheus/common log for logging instead of https://github.com/golang/glog.
+* Use https://github.com/prometheus/common version to build with version information.
+* Use https://github.com/prometheus/promu for building.
```
arvenil: 👍

collector/mongos/server_status_test.go
```diff
@@ -44,25 +36,13 @@ func Test_ParserServerStatus(t *testing.T) {
...
 		t.Error("Connections group was not loaded")
 	}
-
-	if serverStatus.Locks == nil {
-		t.Error("Locks group was not loaded")
-	}
-
-	if serverStatus.Metrics.Document.Deleted != 45726 {
-		t.Error("Metrics group was not loaded correctly")
-	}
```
arvenil: Why those were removed?

mongodb_exporter.go
```diff
@@ -11,137 +12,156 @@ import (
...
-	password string
+	webAuth
+	handler http.HandlerFunc
 }
 
 func (h *basicAuthHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	user, password, ok := r.BasicAuth()
-	if !ok || password != h.password || user != h.user {
-		w.Header().Set("WWW-Authenticate", "Basic realm=\"metrics\"")
+	username, password, _ := r.BasicAuth()
```
arvenil: Why did you skip check for `ok`?

CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
+# Changelog
+
+## v0.2.0 (2017-06-16)
+
+* Default listen port changed to 9216.
```
AlekSi: > Why

It was `9104`, that's mysqld_exporter port. Now we have our own. https://github.com/prometheus/prometheus/wiki/Default-port-allocations

> Where

EXPOSEd port in Dockerfile and default value of `-web.listen-address` flag.

> does it matter for PMM

Yes.

mongodb_exporter.go
```diff
@@ -11,137 +12,156 @@ import (
...
+	usernameOk := subtle.ConstantTimeCompare([]byte(h.Username), []byte(username)) == 1
+	passwordOk := subtle.ConstantTimeCompare([]byte(h.Password), []byte(password)) == 1
```
arvenil: What's the point of `ConstantTimeCompare` here? Isn't user and pass here in plain-text?

CHANGELOG.md
arvenil: I thought we always pass `-web.listen-address` flag?
```
# sudo systemctl show pmm-mongodb-metrics-42003.service | grep ExecStart
ExecStart={ path=/bin/sh ; argv[]=/bin/sh -c /usr/local/percona/pmm-client/mongodb_exporter -web.listen-address=163.172.28.90:42003 -web.auth-file=/usr/local/percona/pmm-client/pmm.yml -web.ssl-cert-file=/usr/local/percona/pmm-client/server.crt -web.ssl-key-file=/usr/local/percona/pmm-client/server.key >> /var/log/pmm-mongodb-metrics-42003.log 2>&1 ; ignore_errors=no ; start_time=[n/a] ; stop_time=[n/a] ; pid=0 ; code=(null) ; status=0/0 }
```

AlekSi: @delgod @EvgeniyPatlan please check this ^

CHANGELOG.md
AlekSi: I think it was @delgod who asked to log less. 😄 I changed logging to default levels for other Prometheus exporters. There are still should be enough information to track errors.

collector/mongos/server_status_test.go
AlekSi: Those metrics are no longer present, tests were not compiling.

CHANGELOG.md
arvenil: 👍

collector/mongos/server_status_test.go
arvenil: 👍

mongodb_exporter.go
AlekSi: If not ok, `username` and `password` will be empty strings which are good enough for us.

mongodb_exporter.go
AlekSi: To prevent timing attack.

arvenil: It doesn't seem to me like a correct usage of `ConstantTimeCompare` that's why I ask. Did you get this idea from somewhere? IMO `ConstantTimeCompare` is for slices with the same, constant length. If a and b has different length, it's a bug; if a has different length for every request (e.g. multiple users with different passwords) - it's a bug.

https://github.com/golang/go/issues/18936#issuecomment-277429198

arvenil: After re-thinking, makes sense when user hits the same password length so 👍

PR #77, opened 2017-08-18
adding databases and collections stats collection

DeathBorn:
from both mongos and mongod side.

### DIFF
--- a/collector/mongod/collections_status.go
+++ b/collector/mongod/collections_status.go
@@ -1,0 +1,128 @@
+package collector_mongod
+
+import (
+	"github.com/prometheus/common/log"
+	"github.com/prometheus/client_golang/prometheus"
+	"gopkg.in/mgo.v2"
+	"gopkg.in/mgo.v2/bson"
+)
+
+var (
+	collectionSize = prometheus.NewGaugeVec(prometheus.GaugeOpts{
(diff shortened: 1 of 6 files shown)


dbmurphy:
This could be rather expensive, and might be a bad item without a way to limit it. Some people have thousands and thousands of names space which will make Prometheus unusable.

timvaillancourt:
I agree with @dbmurphy, this is very expensive. In the dcu/mongodb_exporter this was implemented as an optional flag if I remember correctly. This might be the safer option.

AlekSi:
May you please a PR against `develop` branch?

DeathBorn:
Good insight . I'll rework this accordingly.

DeathBorn:
@AlekSi good.
However, I noticed one bad thing happening with collections metrics. Time to time, I get this error ```was collected before with the same name and label values``` 
. So far, I could not find the reason, why it is happening. it looks like as if exporter can't work with multiple scrapers. Any idea? it happens only with collections metrics.

iamanv:
@DeathBorn How did you run the exporter to be able to export DB metrics for mongod? Running the dbstat command from the exporter always returns an empty document for a dbname. https://github.com/percona/mongodb_exporter/issues/285
Any pointers?

PR #81, opened 2017-09-06
Document MONGODB-X509 auth

atward:
I was concerned that I needed to rewrite the authentication code for mongodb_exporter - but the mechanism can be provided by URI. This documents this for future users.


AlekSi:
Thanks for your contribution! And sorry for a great delay.

PR #87, opened 2018-01-13
 Support integration tests against non-locahost db

bz2:
Use connection string from `TEST_MONGODB_URL` envvar which defaults to `localhost:27017` still.

This makes testing against my local minikube-deployed mongodb easier. Using the `TEST_` prefix from past experience of people running tests against dbs where they actually cared about the data in by mistake when using just `MONGODB_URL`.

Also drive-by ignore my editor's temp files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
+# editor temp files
+*.sw[nop]
+
+# build artefacts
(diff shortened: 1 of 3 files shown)


AlekSi:
Merged. Thank you!

PR #86, opened 2018-01-13
PMM-2261 Unify common mongod and mongos server metrics

bz2:
New collector_common package for metrics that are the same in both collector_mongod and collector_mongos packages.

The labels of these metrics are now prefixed with just 'mongodb_' rather than 'mongodb_mongo[ds]_'.

Note that because bson only supports ",inline" for maps and structs, the common ServerStatus is embedded by-value not by-pointer.

This PR is about removing duplicated code rather than getting the semantics exactly right. At least MemStats should be split as it has some old storage engine specific values. I'm not keen on including "process" in the metric name, but it would be possible to keep the current layout if that's really needed.

PR #88, opened 2018-01-13
Export mongodb version as a metric

bz2:
Expose as a gauge with the version as a label and a value of 1.

Follows this scheme for doing software versions:
https://www.robustperception.io/exposing-the-software-version-to-prometheus/

Will conflict with https://github.com/percona/mongodb_exporter/pull/86 but I can resolve that.


AlekSi:
Merged, thanks!

PR #94, opened 2018-03-23
PMM-2529: Enable collection of table top metrics

bobera:
Copied from @dcu Top Metrics per collection metrics. Disabled by default, to enable use -collect.topmetrics.


timvaillancourt:
@bobera / @arvenil : Thanks for the contribution.

Adding this to PMM would mean that an additional 18 metrics would be added to Prometheus for every collection in the database server, an unbounded number... If you had 100 collections this 1800 additional metrics to the several-hundred base PMM metrics.

As Prometheus best practices strongly advise against using metrics with an unbounded number of labels we need to be very careful of the costs to the PMM (Prometheus) server and also consider if fetching something like this every N x seconds is valuable. Perhaps every N x minutes would be fine, etc.

PR #97, opened 2018-04-11
PMM-2528  Individual index usage stats and index sizes added

martinhoefling:
This pull requests adds sizes of individual indexes and individual index usage statistics. The latter can be activated via a separate flag, as it requires another query.


arvenil:
@roman-vynar @timvaillancourt would we find this useful to be included in PMM dashboards?

collector/mongod/collections_status.go
```diff
@@ -74,15 +80,24 @@ func (collStatList *CollectionStatList) Export(ch chan<- prometheus.Metric) {
 		collectionObjectCount.With(ls).Set(float64(member.Count))
 		collectionAvgObjSize.With(ls).Set(float64(member.AvgObjSize))
 		collectionStorageSize.With(ls).Set(float64(member.StorageSize))
-		collectionIndexes.With(ls).Set(float64(member.Indexes))
+		collectionIndexes.With(ls).Set(float64(len(member.IndexSizes)))
```
arvenil: @martinhoefling Could you describe what was wrong here?

arvenil:
@timvaillancourt Thanks for really good points. Since this is disabled by default, could it be acceptable for now as part of mongodb_exporter but not part of PMM?

collector/mongod/collections_status.go
martinhoefling: I had to map `bson:"indexSizes,omitempty"` in `CollectionStatus` struct above to a different type to make it available for a per index size map. Therefore, i need to figure out the number of total indexes slightly different.

martinhoefling:
@timvaillancourt thx for your (valid) concerns 1-3
1. as said before, detailed collection index usage statistics are disabled by default. We're running in production ~20 collections, ~10 indexes each using a scraping interval of 15s with no notable performance impact. But as you said, this might vary a lot depending on the concrete collection and index sizes.
2. Sorry, can't tell you much as I do not know PMM
3. Depends a bit on the definition of "unbound" but I agree that for 100 collections with 100 indexes each that would create 10k metrics just for the individual index sizes. One option would be to make the output disableable as well.

What do you think?

dbmurphy:
I would advocate maybe  a middle ground, off by default, reports top 10 collections by size and its index when enable, unless someone change  -namespace-stats-limit  to something other than 10

collector/mongod/index_usage.go
```diff
@@ -0,0 +1,81 @@
+package collector_mongod
```
golangci: don't use an underscore in package name

golangci: don't use an underscore in package name

PR #100, opened 2018-04-12
PMM-2530: Add timeout flags

unguiculus:
Adds the following flags:

* -mongodb.socket-timeout
* -mongodb.sync-timeout

I left the defaults as they were, but I think it might be a good idea to change them as well. The MongoDB driver has a default for the sync timeout of 7 seconds ([session.go#L1706](https://github.com/percona/mongodb_exporter/blob/develop/vendor/gopkg.in/mgo.v2/session.go#L1706)) and a default value for the socket timeout of one minute ([session.go#L1716](https://github.com/percona/mongodb_exporter/blob/develop/vendor/gopkg.in/mgo.v2/session.go#L1716)) whereas the exporter uses a much longer default for the sync timeout of one minute and an extremely short default of 3 seconds for the socket timeout.

We've found that the default socket timeout is way too short when you have a recovering replica. It may take quite a bit to connect. You want to make sure to get your metrics, so I'd rather set the timeout higher.

### DIFF
--- a/mongodb_exporter.go
+++ b/mongodb_exporter.go
@@ -19,6 +19,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/percona/exporter_shared"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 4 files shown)


unguiculus:
Any chance this gets merged?

mongodb_exporter.go
```diff
@@ -62,6 +63,9 @@ var (
 	maxConnectionsF               = flag.Int("mongodb.max-connections", 1, "Max number of pooled connections to the database.")
 	testF                         = flag.Bool("test", false, "Check MongoDB connection, print buildInfo() information and exit.")
 
+	socketTimeoutF = flag.Int64("mongodb.socket-timeout", int64(3*time.Second), "Amount of time in ms to wait for a non-responding socket to the database before it is forcefully closed")
+	syncTimeoutF   = flag.Int64("mongodb.sync-timeout", int64(1*time.Minute), "Amount of time in ms an operation with this session will wait before returning an error in case a connection to a usable server can't be established")
```
arvenil: There is [flag.Duration](https://golang.org/pkg/flag/#Duration) for this purpose.
Example https://github.com/percona/mysqld_exporter/blob/50377ef0863147751e3057fa95b263ed396ad2f4/mysqld_exporter.go#L82

timvaillancourt:
@unguiculus Thanks for the contribution. Approved by me

mongodb_exporter.go
timvaillancourt: Good idea!

unguiculus:
@arvenil @timvaillancourt Changed it to `flag.Duration`. Thanks.

PR #102, opened 2018-05-03
ARM64-specific fixes.

akopytov:
2 portability issues in Makefile:

1. Race detector is only supported on amd64, so '-race' cannot be
used other architectures
2. 'uname -m' reportes 'aarch64' on arm64 CPUs, but the corresponding
value accepts in GOARCH is 'arm64'.


arvenil:
@akopytov Thanks for your contribution!

PR #106, opened 2018-05-21
PMM-2620: Fix balancerIsEnabled & balancerChunksBalanced values

jmsantorum:
After checking the code I figure out that the values for **balancerIsEnabled** and **balancerChunksBalanced** were exchanged


arvenil:
@jmsantorum Thanks for really nice catch!

PR #103, opened 2018-05-17
PMM-2591: v0.6.0

arvenil:
[v0.6.0](https://github.com/percona/mongodb_exporter/releases/edit/untagged-631e22276bb445840e1e)
* Add timeout flags #100
* Enable collection of table top metrics #94
* Individual index usage stats and index sizes added #97
* Support &ssl=true #105 #90
* Fix balancerIsEnabled & balancerChunksBalanced values  #106
* Add required testify dependency #107

PR #107, opened 2018-06-01
Add required testify dependency

RubenHoms:
I wanted to `make` this project but was unable to due to a missing dependency. I ran a `dep ensure` and it now `make`'s it without problems.

I couldn't open an issue so I'll ask for it here. I see that you are using the `.promu.yml` file to be able to crossbuild releases which is great. What I'm missing though is these built binaries being put in the actual releases, which would make this project that much more helpful for installing on a server.

For example, I use SaltStack as a configuration management tool and would rather download a binary then to have to install all the dependencies and then build it myself. This is error prone and time-consuming which is unnecessary since you've got the tools already set up to release these binaries.

Would you accept a pull request which would improve the `.promu.yml` build a bit further so you can simply `promu crossbuild` and then upload the binaries as you're making a release? I see you're building with travis, but I'm not sure how to automatically upload releases after a git tag has been made as I'm not too familiar with the system. 

Here's some examples for how I do releases for exporter projects (I basically copied the practice from the `node-exporter` repo.):
* [MaxScale exporter](https://github.com/skord/maxscale_exporter/releases)
* [OpenSIPs exporter](https://github.com/VoIPGRID/opensips_exporter/releases)

Cheers! And thanks for the work put into this exporter :+1:

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -50,6 +50,17 @@
   revision = "7d9e9cc0256a3fbb559aa2f3fb9d8d37e3bd1bf5"
 
 [[projects]]
+  name = "github.com/percona/mongodb_exporter"
+  packages = [
+    "collector",
+    "collector/mongod",
+    "collector/mongos",
+    "shared"
+  ]
+  revision = "8136fdd3628a63985edf3b9a30ceec757638ff53"
+  version = "v0.5.0"
+
+[[projects]]
   name = "github.com/pmezard/go-difflib"
   packages = ["difflib"]
   revision = "792786c7400a136282c1664665ae0a8db921c6c2"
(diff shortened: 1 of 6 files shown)


Gopkg.lock
```diff
@@ -49,6 +49,17 @@
...
+  name = "github.com/percona/mongodb_exporter"
+  packages = [
+    "collector",
+    "collector/mongod",
+    "collector/mongos",
+    "shared"
+  ]
+  revision = "8136fdd3628a63985edf3b9a30ceec757638ff53"
+  version = "v0.5.0"
+
```
arvenil: Why it depends on itself?

vendor/github.com/stretchr/testify/require/doc.go
```diff
@@ -0,0 +1,28 @@
+// Package require implements the same assertions as the `assert` package but
```
arvenil: Thanks, I just noticed this missing dependency too and it will be added in v0.6.0 https://github.com/percona/mongodb_exporter/pull/103

PR #113, opened 2018-07-10
Fix typo. Make the environment variable MONGODB_URL not MONGODB_URI

bramblek1:
Despite the flag being called mongodb.uri, the documentation and the default
value of mongdb.uri refer to the env var MONGODB_URL

When overriding this prior to creating a test connection , the value of os.GetEnv("MONGODB_URI") is used to override. 

In the scenario where you've got mongodb_exporter being launched with a flag for the mongdb.uri AND you're overriding that with MONGODB_URL in the environment. Results are confusing.

### DIFF
--- a/mongodb_exporter.go
+++ b/mongodb_exporter.go
@@ -84,7 +84,7 @@ func main() {
 	}
 	flag.Parse()
 
-	uri := os.Getenv("MONGODB_URI")
+	uri := os.Getenv("MONGODB_URL")
 	if uri != "" {
 		uriF = &uri
 	}


arvenil:
Hi @bramblek1 
Thank you for your contribution. This sadly goes on hold. The problem is this exporter is part of [PMM](https://www.percona.com/software/database-tools/percona-monitoring-and-management) and PMM uses `MONGODB_URI` instead. Looks like old error we need to fix but it will need to wait until we fix it also in PMM, which may take some time as we need to keep also backward compatibility. In the meantime you can use `MONGODB_URI` instead. I've also filled upstream bug report to clarify this https://github.com/dcu/mongodb_exporter/issues/108.

AlekSi:
We will use `URI` instead of `URL` everywhere.

PR #114, opened 2018-08-06
Add Replication Lag metrics for secondaries.

anas-aso:
I tried running the tests, but they are already broken for master branch.


timvaillancourt:
FYI, this data can already be gathered by subtracting the optimes of the Secondary and Primary. This is what Percona PMM does today to measure replication lag.

anas-aso:
@timvaillancourt Thanks for your reply. I agree, but I think it's useful to have it available out of the box since it's an interesting metrics (at least for our use case). 
I used the way you mentioned before, but wasn't "comfortable" and missed a couple of labels I needed due to the aggregation I had to do.

DeathBorn:
@timvaillancourt there is something not nice regarding replication lag when calculation happens via queries through prometheus server. nodes are not scraped at the same time and that creates ugly metrics. Calculating lag from the node perspective might be helpful here.

timvaillancourt:
This PR looks good to me, but I suggest the lag is calculated more accurately.

Values in 'replSetGetStatus' outputs are updated on replica set heartbeats, which have some delay. For this lag metric to accurate, the staleness of the oplog timestamps must be taken into consideration.

The "staleness" of the oplog timestamp can be calculated by subtracting a replset member's 'lastHeartBeatRecv' (https://docs.mongodb.com/manual/reference/command/replSetGetStatus/#replSetGetStatus.members[n].lastHeartbeatRecv) from the top-level 'date' (https://docs.mongodb.com/manual/reference/command/replSetGetStatus/#replSetGetStatus.date) field. The difference between these two values is usually called "operational lag" and this should be subtracted from the result.

> replSetGetStatus.members[n].lastHeartbeatRecv
An ISODate formatted date and time that reflects the last time the server that processed the replSetGetStatus command received a heartbeat request from this member (members[n]). Compare this value to the values of the date and lastHeartBeat fields to track latency between these replica set members.

So, if for example: there is 3 seconds of replication lag and there is .5 seconds of "operational lag", the lag really is 2.5 seconds. Exposing "operational lag" as a new metric would probably be the most-flexible way to do this although this would require a subtraction of the two lag values in your Prometheus query.

geekdave:
@timvaillancourt While waiting for this PR, I tried out your suggestion:

> FYI, this data can already be gathered by subtracting the optimes of the Secondary and Primary (exported by mongodb_exporter).

I used this query in Grafana:

```
sum(mongodb_mongod_replset_member_optime_date{instance="mongo-primary", state="PRIMARY"}) - sum(mongodb_mongod_replset_member_optime_date{instance="mongo-primary", name=~"mongo-secondary", state="SECONDARY"}) 
```

But it creates a rather spikey graph, with unexpected large and brief spikes.  I would expect replication lag to be something that shows a ramp-up, rather than a spike.  Is there a way that you would model this query differently?  I am guessing the way I have it now, perhaps the data points are misaligned.

Thanks!

anas-aso:
Hi @timvaillancourt. Thanks for having a second look and sorry.
I tried what you mentioned but it didn't work as `member.LastHeartbeatRecv` is always `nil` for secondaries and the replicationLag is defined for Secondaries only.

timvaillancourt:
@anas-aso if the replSetGetStatus is retrieved from a Secondary the operational lag should be subtracted from the Primary member. The replSetGetStatus output from a Secondary will return a Primary state (oplog time, etc) that is as stale as the last heartbeat received from it.

anas-aso:
@timvaillancourt please have another look. I have added the operational lag as a separated metric. I think this way it's better for more flexibility.

timvaillancourt:
@anas-aso: this is fantastic. I agree that exposing a 2nd metric makes more sense.

Approved, thanks for the contribution!

anas-aso:
@timvaillancourt is there something missing to merge this ? We would like to switch to this repo instead of my fork.

timvaillancourt:
@anas-aso: @AlekSi or @rnovikovP handle merging/releasing, let's wait for their comment

anas-aso:
@AlekSi @rnovikovP can you please have a look at this PR ?

tetianakravchenko:
@AlekSi @rnovikovP is it planned to make a release that includes those changes?

PR #121, opened 2018-11-14
Fix for broken labels on index_usage metrics

akira-kurogane:
The index stats lacked db as a label. I.e. a "foo" collection's stats from databases "x" and "y" would have been indistinct, or one overwriting the other. This fixes that.

### DIFF
--- a/collector/mongod/index_usage.go
+++ b/collector/mongod/index_usage.go
@@ -12,7 +12,7 @@ var (
 		Namespace: Namespace,
 		Name:      "index_usage_count",
 		Help:      "Contains a usage count of each index",
-	}, []string{"collection", "index"})
+	}, []string{"collection", "db", "index"})
 )
 
 // IndexStatsList represents index usage information
@@ -24,6 +24,7 @@ type IndexStatsList struct {
 type IndexUsageStats struct {
 	Name       string         `bson:"name"`
 	Accesses   IndexUsageInfo `bson:"accesses"`
+	Database   string
 	Collection string
 }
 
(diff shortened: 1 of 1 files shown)


AlekSi:
>  I.e. a "foo" collection's stats from databases "x" and "y" would have been indistinct, or one overwriting the other.

Aren't current code adds database name as a prefix for collection name?

akira-kurogane:
Ah, so it does! Maybe I removed db name by mistake while debugging why I wasn't getting any results, which was ultimately the issue in pull request #122.

I thought "db + . +  collectionName" was my temporary work, and then I polished that some more by making db a separate label as is done for collector/mongod/collections_status.go.

If you like the separated db label, then this patch. If not, no need for this patch then I guess.

rnovikovP:
my 2c:  to aggregate data  by labels will be simpler than "cut" it by "."
So even if this will require to do some dashboards changes - it's better to have these parameters separated

klynch:
+1 on merging this!

klynch:
@AlekSi Any chance this can get merged and we can get a release? I'm running into this issue and may end up cutting a release from this branch.

AlekSi:
> If you like the separated db label, then this patch. If not, no need for this patch then I guess.

Is there a case where you want to aggregate by collection name, but not by database name / "full" collection name?

AlekSi:
Ok, I will merge this. Aggregating by database name makes sense.

AlekSi:
Thank you!

akira-kurogane:
My pleasure.

PR #122, opened 2018-11-15
PMM-4541 PMM-4906 Exclude system collections from collStats and indexStats collecting

akira-kurogane:
The system collections (those in admin, config or local db, or those prefixed with "system." such as system.profile, system.users) don't need to be monitored.
A) They're correctly indexed and what use they get is necessary, so there's nothing to optimize on them.
B) Permissions for these collections are tricky. They are an exception to the permissions granted by built-in roles. In reality people who want to use these collection and index-level stats will mostly give up and use "root" or something like that if the system collections are included in these mongodb_exporter collection routines.

One exception is the _local.oplog.rs_ collection, but the metrics.repl.* metrics in serverStatus are already covering that.

This patch changes collections_status.go and index_usage.go so they wont iterate over any system.* collections or anything in the "local" or "admin" or "config" dbs. I.e. it will only iterate user-created collections.

----

In case you land here from websearch please note even with this patch using the -collect.collection and -collect.indexusage options requires the DBA grants some custom roles you don't get with the built-in _clusterMonitor_ role. Actions "listCollections", "collStats" and "indexStats" need to be packed in a custom role that is granted that to the login the monitoring agent uses.

```
db.createRole({
  role: "my_collstats_monitoring_role", 
  privileges: [ 
    {resource: {db: "", collection: ""}, 
     actions: ["listCollections", "collStats", "indexStats" ] } 
  ], 
  roles: []
})
db.grantRolesToUser(
  "my_monitoring_user", 
  [ /*built-in role*/ "clusterMonitor", 
    /*this custom role*/ "my_collstats_monitoring_role"]
)
```


AlekSi:
I don't know if we want that by default. I would suggest making a new flag to skip those collections, with default value `false`.

akira-kurogane:
Is there anything I can do to help with the review of this? As far as I understand it's awaiting a human reviewer, but if it's the codecov auto-check is still considered a blocker what can I do about that? The explanation points to flags but I don't find anything in the codecov UI that helps with that.

AlekSi:
Please ignore codecov for now. Let's add a flag for skipping those collections, but do not skip them by default.

And we are sorry for being slow in reviewing PRs. Most of us are currently busy with PMM 2.0 development. But we plan to address ongoing exporters work soon.

AlekSi:
Ok, logging error and continuing sounds good.

akira-kurogane:
OK, I'll close this one.

akira-kurogane:
Other patch created as #128.

akira-kurogane:
Reopening this.

@MajorBreakfast has found that using the --collect.collection and --collect.indexusage options is causing log bloat in the **mongod log**, even after pull request #128 is merged. It is true, the patch of pull request 128 only stops log bloat in the mongodb_exporter log, it does not stop it from continuing in the mongod logs.

The inbuilt role for monitoring ("clusterMonitor") does not provide execution privilege on these collections, and I don't think they ever will because MongoDB developers and support engineers themselves would never run collStats and $indexStats metrics collection on them because they are not the cause of load issues.

The only way to avoid this log bloat issue is to not attempt to run collStats or $indexStats against these collections in the first place.

(For what this patch does please see the original description.)

akira-kurogane:
Reopening this.

@MajorBreakfast has found that using the --collect.collection and --collect.indexusage options is causing log bloat in the mongod log, even after pull request #128 is merged. It is true, the patch of pull request 128 only stops log bloat in the mongodb_exporter log, it does not stop it from continuing in the mongod logs.

The inbuilt role for monitoring ("clusterMonitor") does not provide execution privilege on these collections, and I don't think they ever will because MongoDB developers and support engineers themselves would never run collStats and $indexStats metrics collection on them because they are not the cause of load issues.

The only way to avoid this log bloat issue is to not attempt to run collStats or $indexStats against these collections in the first place.

(For what this patch does please see the original description.)

akira-kurogane:
An existing JIRA ticket is: https://jira.percona.com/browse/PMM-4541

vsetka:
So, when will this be merged? Any ETA?

sonujatav35:
Hi Akira,

I am unable to get the data by --collect.indexusage flag. Even-though i just added extra action to monitoring user along with clusterMonitor role.

akira-kurogane:
Hi @sonujatav35 in the far-away land of Futako-Tamagawa!

It could be a permissions issue on system collections (like this pull request talks about) and if so it wont work (it will try every time, but do nothing once it hits the first collection it can't run $indexStats on.)

I'll ping you 1-to-1 in another channel, lets see if we can get to the bottom of it.

AlekSi:
#177 included all fixes from that PR and was merged. It will be a part of 0.11.0 release. Closing.

PR #126, opened 2018-12-19
Fix index_usage_count labels "db" and "coll"

robinro:
Currently the metric `index_usage_count` only has a label `collection`, which is always empty.
This PR replaces this label with labels `db` and `coll` that match the labels in other metrics.


AlekSi:
We merged #121 which is doing the same, but a bit differently.

PR #130, opened 2019-02-15
PMM-3512 Switch flags to kingpin

DexterHD:


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- PMM-3512: Switched to [kingpin](https://github.com/alecthomas/kingpin) library.
```
AlekSi: It probably makes sense to use links for public Jira tickets.

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+This is **BREAKING CHANGE** because kingpin uses `--` neither `-` for long flags, so be careful when update.
```
AlekSi:
```suggestion
This is a **BREAKING CHANGE** because kingpin uses `--` instead of `-` for long flags, so be careful when updating.
```

mongodb_exporter.go
```diff
@@ -42,47 +41,42 @@ func defaultMongoDBURI() string {
...
-
-	collectDatabaseF   = flag.Bool("collect.database", false, "Enable collection of Database metrics")
-	collectCollectionF = flag.Bool("collect.collection", false, "Enable collection of Collection metrics")
-	collectTopF        = flag.Bool("collect.topmetrics", false, "Enable collection of table top metrics")
-	collectIndexUsageF = flag.Bool("collect.indexusage", false, "Enable collection of per index usage stats")
-
-	uriF     = flag.String("mongodb.uri", defaultMongoDBURI(), "MongoDB URI, format: [mongodb://][user:pass@]host1[:port1][,host2[:port2],...][/database][?options]")
-	tlsF     = flag.Bool("mongodb.tls", false, "Enable tls connection with mongo server")
-	tlsCertF = flag.String("mongodb.tls-cert", "", "Path to PEM file that contains the certificate (and optionally also the decrypted private key in PEM format).\n"+
+	versionF       = kingpin.Flag("version", "Print version information and exit.").Short('v').Bool()
```
AlekSi: Short `-v` is often used for verbose output, so let's drop it there.

mongodb_exporter.go
```diff
@@ -42,47 +41,42 @@ func defaultMongoDBURI() string {
...
+	listenAddressF = kingpin.Flag("web.listen-address", "Address to listen on for web interface and telemetry.").Default(":9216").String()
+	metricsPathF   = kingpin.Flag("web.metrics-path", "Path under which to expose metrics.").Default("/metrics").String()
```
AlekSi: Canonical name is `web.telemetry-path`: https://github.com/prometheus/node_exporter/blob/f028b816152f6d5650ca2cd707e45cda7333fdc1/node_exporter.go#L138

mongodb_exporter.go
```diff
@@ -42,47 +41,42 @@ func defaultMongoDBURI() string {
...
 func main() {
-	flag.Usage = func() {
-		fmt.Fprintf(os.Stderr, "%s %s exports various MongoDB metrics in Prometheus format.\n", os.Args[0], version.Version)
-		fmt.Fprintf(os.Stderr, "Usage: %s [flags]\n\n", os.Args[0])
-		fmt.Fprintf(os.Stderr, "Flags:\n")
-		flag.PrintDefaults()
-	}
-	flag.Parse()
+	kingpin.CommandLine.Help = fmt.Sprintf("%s %s exports various MongoDB metrics in Prometheus format.\n", os.Args[0], version.Version)
+	kingpin.Parse()
```
AlekSi:
```go
kingpin.HelpFlag.Short('h')
```

Plus add `kingpin.Version(version.Version)` or something like that.

testdata/mongodb_exporter.testFlagHelp.golden
```diff
@@ -1,53 +1,57 @@
...
+                                 closed.
+                                 
+                                   Valid time units are 'ns', 'us' (or 'µs'), 'ms', 's', 'm', 'h'.
+      --mongodb.sync-timeout=1m  Amount of time an operation with this session
+                                 will wait before returning an error in case
+                                 
+                                   a connection to a usable server can't be established.
+                                   Valid time units are 'ns', 'us' (or 'µs'), 'ms', 's', 'm', 'h'.
+      --groups.enabled=""        Currently ignored
+
```
AlekSi: I honestly do not know if we need that test at all…

DexterHD: Why not? It works fine and it helps to prevent accidental flags change.

mongodb_exporter.go
DexterHD: I'll add short help flag, but there is no reason to add `kingpin.Version(version.Version)` cause now we have extended version info right here https://github.com/percona/mongodb_exporter/blob/5fa2ea81292eb750161ca79eae66fce5a168e668/mongodb_exporter.go#L110
If I use it, we will see only `X.Y.Z` in output and there is no way to add custom version output.

AlekSi: That `if` should be removed – kingpin will handle that flag itself.
In the future, we will switch to https://github.com/percona/pmm/blob/PMM-2.0/version/release.go like that: https://github.com/percona/pmm-agent/pull/12/commits/9e5b9a1ead40d76afa7651e65ce641d71bdf04e8

testdata/mongodb_exporter.testFlagHelp.golden
AlekSi: Because you don't actually check that – you just copy the actual output to the golden file. But that's up to you.

mongodb_exporter.go
AlekSi: See https://jira.percona.com/browse/PMM-2567, https://jira.percona.com/browse/PMM-1168, and https://jira.percona.com/browse/PMM-3513.

mongodb_exporter.go
```diff
@@ -42,47 +41,43 @@ func defaultMongoDBURI() string {
...
-		fmt.Fprintf(os.Stderr, "Usage: %s [flags]\n\n", os.Args[0])
-		fmt.Fprintf(os.Stderr, "Flags:\n")
-		flag.PrintDefaults()
-	}
-	flag.Parse()
+	kingpin.HelpFlag.Short('h')
+	kingpin.CommandLine.Help = fmt.Sprintf("%s %s exports various MongoDB metrics in Prometheus format.\n", os.Args[0], version.Version)
+	kingpin.Parse()
 
 	uri := os.Getenv("MONGODB_URI")
```
AlekSi: btw, can move that envvar to kingpin flag definition?

DexterHD: Fixed

mongodb_exporter.go
DexterHD: I thought `version` needs to bee, like `0.6.3` in context of `kingpin.Application`. It was small misunderstanding, so I'll remove it.

testdata/mongodb_exporter.testFlagHelp.golden
DexterHD: I'll make a note and try to figure out how to improve the test in future, but I'm going to keep it for now.
Btw it's hard to change the `.golden` file by hand as output length of the string may differ between your terminal and test output. So you should run `go test ./. -update` manually to change the file. It you forget to do this, tests will fail.

DexterHD:
@AlekSi All remarks was fixed. Can I merge this?

PR #133, opened 2019-02-25
PMM-3524: Add information about missing metrics in test.

DexterHD:
From this moment we can see which metrics wasn't described and/or collected when test fails.

### DIFF
--- a/collector/mongodb_collector_test.go
+++ b/collector/mongodb_collector_test.go
@@ -15,7 +15,6 @@
 package collector
 
 import (
-	"fmt"
 	"os"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


collector/mongodb_collector_test.go
```diff
@@ -55,22 +55,31 @@ func TestCollector(t *testing.T) {
 		close(metricCh)
 	}()
 
-	var descs int
-	for range descCh {
-		descs++
+	var descriptors = make(map[string]struct{})
```
AlekSi: `:=`

collector/mongodb_collector_test.go
```diff
@@ -55,22 +55,31 @@ func TestCollector(t *testing.T) {
...
 	}
 
-	var metrics int
+	var metricsCount int
 	var versionInfoFound bool
 	for m := range metricCh {
+
+		if _, ok := descriptors[m.Desc().String()]; ok {
+			delete(descriptors, m.Desc().String())
+		}
```
AlekSi: `assert.True(t, ok)` with a good message.

DexterHD: I'll print in after the cycle as 2 metrics (with different labels) have same descriptor. and if I do it here we will duplicate output when

AlekSi: Cool. Is it a real error message? That one seems easy to fix.

DexterHD: > Cool. Is it a real error message? That one seems easy to fix.

For now it's just some imitation of an error. But when it happens in CI, it will be real and we can fix it.

PR #141, opened 2019-06-06
PMM-3511 Migrate to MongoDB official driver.

DexterHD:
## Brief

This PR represents a migration from [mgo.v2](https://github.com/globalsign/mgo) to the official [mongo-go-driver](https://github.com/mongodb/mongo-go-driver).

As those drivers have some [conceptual differences](https://www.mongodb.com/blog/post/go-migration-guide), here are some details about migration that should be discussed:

## Changes due to replacing `*mgo.Session` to `*mongo.Client`:

- [DialInfo.FailFast](https://godoc.org/github.com/globalsign/mgo#DialInfo) was removed as it hasn't analog in the new driver.
- `session.SetMode(mgo.Eventual, true)` there is no [mgo.Eventual](https://godoc.org/github.com/globalsign/mgo#Mode) mode in official driver. So it was changed to [readpref.Nearest](https://godoc.org/go.mongodb.org/mongo-driver/mongo/readpref#Mode) mode.
- `session.SetPrefetch(0.00)` was removed, as it's not available in the MongoDB Go driver. 
- `session.SetSyncTimeout(opts.SyncTimeout)` there is no analog of this option in new driver. At this moment I use `opts.SyncTimeout` in [ClientOptions.SetConnectTimeout()](https://godoc.org/go.mongodb.org/mongo-driver/mongo/options#ClientOptions.SetConnectTimeout) function, but I'm not sure it's good choise. Also, it used as default timeout in the `ContextWithTimeout`, when a connection is creating.
- [dialInfo.DialServer](https://github.com/percona/mongodb_exporter/blob/0dcd9d7b7e0c8d58ed44dbca726e1b58c0da2106/shared/connection.go#L116-L129) was replaced to [tlsDialer](https://github.com/percona/mongodb_exporter/blob/e1cd4b627da943af18d2352be5b69ef43a408c67/shared/connection.go#L136-L154).

## `BSON` changes

- ~~`bson.MongoTimestamp` from `gopkg.in/mgo.v2/bson` was replaced to `primitive.Timestamp` from `go.mongodb.org/mongo-driver/bson/primitive`. Because of this, I have changed `mongod.BsonMongoTimestampToUnix()` implementation. [Not sure I have done it right.](https://github.com/percona/mongodb_exporter/pull/141/files#diff-1dfcb33f7cd1c70e6f8f467a3647df8bR77)~~
- `mongod.TopStatus` was tricky to Decode as `{top: 1}` response value `totals` have a field called `note` with not a `TopStatsMap` like struct but `string`. So I have made `TopStatusRaw` intermediate struct to decode a result.

## Other changes

- `AddCodeCommentToQuery()` in `shared/utils.go` was replaced to `GetCallerLocation()` because this is that this function do. It adds information about caller location and places it as a `comment` to `mgo.Query`. Also in the new implementation, this function uses not in all cases, because of new API has no concept like a `mgo.Query` and `query comments` works in a different way. This way is `functional options` and not all API methods have `Comment` option. For more details look at [FindOptions](https://godoc.org/go.mongodb.org/mongo-driver/mongo/options#FindOptions) and [CountOptions](https://godoc.org/go.mongodb.org/mongo-driver/mongo/options#CountOptions) foe example.


AlekSi:
> Because of this, I have changed mongod.BsonMongoTimestampToUnix() implementation. Not sure I have done it right.

Can we write a test using MongoDB in Docker for that?

collector/mongos/collections_status.go
```diff
@@ -100,9 +103,9 @@ var (
 )
 
 // GetCollectionStatList returns stats for a given database
-func GetCollectionStatList(session *mgo.Session) *CollectionStatList {
+func GetCollectionStatList(ctx mongo.SessionContext, client *mongo.Client) *CollectionStatList {
```
AlekSi: All mongod functions were changed to accept `*mongo.Client` instead of `*mgo.Session`. All mongos function were changed to also accept `mongo.SessionContext`. Is there a reason for that? And if yes, is there a reason why `context.TODO()` is used below instead of `ctx`?

collector/mongod/top_status_test.go
```diff
@@ -1,105 +1,42 @@
...
-	if topStats.Remove.Count != 1 {
-		t.Error("Wrong remove operation count value for dummy user collection")
-	}
-
-	if topStats.Commands.Time != 0 {
-		t.Error("Wrong commands operation time value for dummy user collection")
-	}
-	if topStats.Commands.Count != 0 {
-		t.Error("Wrong commands operation count value for dummy user collection")
-	}
```
AlekSi: Can we restore those tests at least partially?

collector/mongos/collections_status.go
```diff
@@ -112,32 +115,50 @@ func GetCollectionStatList(session *mgo.Session) *CollectionStatList {
 		return nil
 	}
 	delete(logSuppressCS, "")
-	for _, dbName := range database_names {
-		collNames, err := session.DB(dbName).CollectionNames()
+	for _, dbName := range dbNames {
+		c, err := client.Database(dbName).ListCollections(ctx, bson.M{}, options.ListCollections().SetNameOnly(true))
+		defer c.Close(context.TODO())
```
percona-csalguero: `defer` inside loop

collector/mongod/top_status_test.go
DexterHD: > Can we restore those tests at least partially?

Could you explain please, why we should restore those tests? I've analyzed it and decide, these checks don't make any sense. Those tests were using fixtures from files and it didn't connect to real MongoDB. All test data was prefetched.

At this point, I have added a connection to real mongo so now it works better than it was. I've removed some check cause numbers can be different from one DB to another. Also, there is no dump of those collections in sources.

I agree we should add real tests but it needs to do some analysis on how to implement it with real different versions of MongoDB and real data. 

I propose to move this in a separate task. This task is only about migrating to the official driver.

collector/mongos/collections_status.go
DexterHD: > All mongos function were changed to also accept mongo.SessionContext. Is there a reason for that?

Yes, here is the reason:

Old implementation:

https://github.com/percona/mongodb_exporter/blob/0dcd9d7b7e0c8d58ed44dbca726e1b58c0da2106/collector/mongodb_collector.go#L238-L240

New implementation:

https://github.com/percona/mongodb_exporter/blob/156801d2b09738a522e620a94cd1ea5af63545cc/collector/mongodb_collector.go#L240-L243

collector/mongod/top_status_test.go
AlekSi: > I've removed some check cause numbers can be different from one DB to another.

At very least, we should check that some numbers are not zero. Otherwise, this test executes code but does not really test results. If the driver or MongoDB itself changed in an unexpected way, this test will still pass.

If we remove tests in this task, then we should add at least some of them back in this task.

collector/mongos/collections_status.go
DexterHD: > That hardly explains it. https://jira.mongodb.org/browse/SERVER-27864 was fixed two years ago in a version that is no longer supported.
> I think we should update all functions to accept only mongo.SessionContext

Then it's not necessary to use `mongo.SessionContext` at all. Especially because [it's not concurrent safe.](https://godoc.org/go.mongodb.org/mongo-driver/mongo#SessionContext).

> and also get a client with ctx.Client().
There is no `ctx.Client()` in driver version `1.0.2` only in `master`.

I propose to remove this context at this time and add context in the next task. Also, I don't like that exporter use global variables through all codebase, it seems we should refactor it.

AlekSi: > I propose to remove this context at this time

Ok, let's accept only Client everywhere.

Agree with other points.

collector/mongod/server_status_test.go
```diff
@@ -15,37 +15,40 @@
...
 
 func TestParserServerStatus(t *testing.T) {
-	data, err := ioutil.ReadFile("../fixtures/server_status.bson")
+
+	serverStatus := &ServerStatus{}
+	client, err := mongo.Connect(context.TODO(), options.Client().ApplyURI("mongodb://localhost:27017"))
 	if err != nil {
 		t.Fatal(err)
 	}
+	defer client.Disconnect(context.TODO())
```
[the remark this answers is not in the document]
DexterHD: In the new driver, you can run a command with a different context and properties. All methods of collection, database, etc. have `context.Context` argument. As I understand from [Usage Examples](https://github.com/mongodb/mongo-go-driver#usage) `connect.Client` doesn't ensure a real connection too, but `RunCommand()` and other methods ensuring it by itself, so you don't need to call `Ping()`. I mean it won't panic.

Also `client.Disconnect()` works [in a different way](https://godoc.org/go.mongodb.org/mongo-driver/mongo#Client.Disconnect) than `session.Close()` in `mgo`.

collector/mongod/top_status_test.go
DexterHD: > If we remove tests in this task, then we should add at least some of them back in this task.

I don't remove the test, only change it to working with a real database and Corecov displays that test coverage was increased.

> At very least, we should check that some numbers are not zero.

Not exactly. This test called: `Test_ParserTopStatus`. So it tests that decoding for this command working well. There is no need to check any numbers in this context. It's enough to check `Decode()` have done and we have a list of collections.

If you think it's not enough, please, explain, why should I improve it and how?

AlekSi: > I don't remove the test

Sorry, the fact is - you did. Red lines show that.

> Corecov displays that test coverage was increased

It shows that more code actually runs during tests. If you call a function in tests without any asserts, it will still improve code coverage number – but it will not test much.

> If you think it's not enough, please, explain, why should I improve it and how?

Because this test will pass if MongoDB decides to rename some metrics we depend on. We want to catch things like that – across different versions.

DexterHD: > Because this test will pass if MongoDB decides to rename some metrics we depend on. We want to catch things like that – across different versions.

But this test doesn't check that in the previous implementation. It checks some fields from the fixture file. It always passed. It was changed only because of the official `bson.Decode()` didn't work with this fake data, so I have removed those fake data and connect it to real mongo with minimal implementation.

AlekSi:
>> Because of this, I have changed mongod.BsonMongoTimestampToUnix() implementation. Not sure I have done it right.

> Can we write a test using MongoDB in Docker for that?

Can we?

DexterHD:
> > Because of this, I have changed mongod.BsonMongoTimestampToUnix() implementation. Not sure I have done it right.
> 
> Can we write a test using MongoDB in Docker for that?

Not need to do this. I read docs one more time,  remove this method and fix the code.

collector/mongod/oplog_status.go
```diff
@@ -104,9 +104,10 @@ func GetOplogTimestamps(session *mgo.Session) (*OplogTimestamps, error) {
 	return oplogTimestamps, err
 }
 
-func GetOplogCollectionStats(session *mgo.Session) (*OplogCollectionStats, error) {
+// GetOplogCollectionStats gets oplog connection stats.
+func GetOplogCollectionStats(client *mongo.Client) (*OplogCollectionStats, error) {
 	results := &OplogCollectionStats{}
-	err := session.DB("local").Run(bson.M{"collStats": "oplog.rs"}, &results)
+	err := client.Database("local").RunCommand(context.TODO(), bson.M{"collStats": "oplog.rs"}).Decode(&results)
```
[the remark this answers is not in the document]
DexterHD: Thanks, @percona-csalguero . I create a new JIRA task, so we can improve this.

PR #143, opened 2019-06-10
PMM-4154 Add log-levels.

DexterHD:


mongodb_exporter.go
```diff
@@ -38,47 +38,49 @@ const (
...
 	// enabledGroupsFlag = flag.String("groups.enabled", "asserts,durability,background_flushing,connections,extra_info,global_lock,index_counters,network,op_counters,op_counters_repl,memory,locks,metrics", "Comma-separated list of groups to use, for more info see: docs.mongodb.org/manual/reference/command/serverStatus/")
-	enabledGroupsFlag = kingpin.Flag("groups.enabled", "Currently ignored").Default("").String()
+	enabledGroupsFlag = app.Flag("groups.enabled", "Currently ignored").Default("").String()
 )
 
 func main() {
-	initVersionInfo()
-	kingpin.Parse()
+	initVersionInfo(app)
+	log.AddFlags(app)
```
AlekSi: Can we use `log.AddFlags(kingpin.CommandLine)` and avoid most of the changes in this PR?

mongodb_exporter.go
```diff
@@ -133,7 +135,8 @@ func main() {
 // to application version and will be printed in all logs.
 // TODO: Refactor after moving version.Info() and version.BuildContext() to https://github.com/percona/exporter_shared
 // See: https://jira.percona.com/browse/PMM-3250 and https://github.com/percona/mongodb_exporter/pull/132#discussion_r262227248
-func initVersionInfo() {
+func initVersionInfo(app *kingpin.Application) {
+
```
AlekSi: Do not add empty lines at the begging of the functions.

mongodb_exporter.go
DexterHD: I prefer to remove a global state when and where it can be removed because it, creates one of the most dangerous indirection points. But ok, as you want.

DexterHD:
@AlekSi ping

AlekSi:
Let's make a patch release with those fixes, but without new driver.

PR #145, opened 2019-06-17
PMM-4168 Update deps.

DexterHD:
Rebased on top of #141.


Gopkg.lock
```diff
@@ -88,27 +112,27 @@
...
   packages = [
     "prometheus",
+    "prometheus/internal",
     "prometheus/promhttp",
   ]
   pruneopts = "NUT"
-  revision = "c5b7fccd204277076155f10851dad72b76a49317"
-  version = "v0.8.0"
+  revision = "2641b987480bca71fb39738eb8c8b0d577cb1d76"
+  version = "v0.9.4"
```
AlekSi: > PMM-4168 Update prometheus/client library to latest version.

The latest version is 1.0.0

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Changed
+- [PMM-4168](https://jira.percona.com/browse/PMM-4168): Updated all dependencies including `github.com/prometheus/golang_client`. 
+As a result a lot of metrics was changed from Counter to Gauge. For more information look at commit [af19190](https://github.com/percona/mongodb_exporter/commit/af1919030e6b9cf3a3cc6da069435f09d49f5099).
```
AlekSi: That commit does not really explain why you changed metric types.
I assume that this is because of Prometheus client removing `Set` method for counters in 0.9.0. But many (most? all?) metrics that were changed to gauges should remain counters for Prometheus to be able to detect resets and for our dashboards to continue working.
I mentioned the proper fix in the other ticket: `NewConstMetric`. See here: https://godoc.org/github.com/prometheus/client_golang/prometheus#hdr-Custom_Collectors_and_constant_Metrics

collector/common/asserts.go
```diff
@@ -19,14 +19,15 @@ import (
...
-	assertsTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
+	assertsTotal = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: Namespace,
 		Name:      "asserts_total",
 		Help:      "The asserts document reports the number of asserts on the database. While assert errors are typically uncommon, if there are non-zero values for the asserts, you should check the log file for the mongod process for more information. In many cases these errors are trivial, but are worth investigating.",
 	}, []string{"type"})
 )
 
 // AssertsStats has the assets metrics
+
```
AlekSi: Please revert that change.

Gopkg.lock
AlekSi: There are new releases of `exporter_shared`: https://github.com/percona/exporter_shared/releases

CHANGELOG.md
DexterHD: > I assume that this is because of Prometheus client removing Set method for counters in 0.9.0.

Yes.

> But many (most? all?) metrics that were changed to gauges should remain counters 

So many, most, or all? How you propose to increment these values without `Set()` method? Should we save previous value after each collect and then subtract it from the next value, or what?

> to be able to detect resets

Could you explain please what did you mean? How we are detecting resets and for what reasons we are detecting it? How does it work?

> and for our dashboards to continue working

Why they will break? Because of metric types changes? Why we can't update our dashboards in this case? 

> I mentioned the proper fix in the other ticket: NewConstMetric.

`NewConstMetric returns a metric with one fixed value that cannot be changed.`
I'm not sure how it can help. Can you give a little bit more information and context, how it can help and with what exactly?

AlekSi: Did you read the link?..

> The Collector example illustrates the use case. You can also look at the source code of the processCollector (mirroring process metrics), the goCollector (mirroring Go metrics), or the expvarCollector (mirroring expvar metrics) as examples that are used in this package itself.

That's what we should do there.

Re counters vs gauges:
* https://prometheus.io/docs/concepts/metric_types/
* https://www.robustperception.io/how-does-a-prometheus-counter-work
* https://prometheus.io/docs/prometheus/latest/querying/functions/#rate

When MongoDB restarts, it resets all metrics. PromQL `rate()` function can detect that for a counter to do the right thing, but not for a gauge. No, we can't fix dashboards if incorrect data is already in.

> How you propose to increment these values without Set() method?

By using `NewConstMetric`. Did you really type that question before finishing reading? Why?

DexterHD: > Did you read the link?.

Yes, I read the link but doesn't see examples at this moment.

> By using NewConstMetric. Did you really type that question before finishing reading? Why?

So I should revert all Metrics I have changed and replace all of them to `NewConstMetric()` with type Counter? Am I understand right?

AlekSi: > Yes, I read the link but doesn't see examples at this moment.

* https://github.com/prometheus/client_golang/blob/master/prometheus/process_collector.go
* https://github.com/prometheus/client_golang/blob/master/prometheus/go_collector.go
* https://github.com/prometheus/client_golang/blob/master/prometheus/expvar_collector.go

You can also search at PMM code and exporters for `NewConstMetric` examples.

> So I should revert all Metrics I have changed and replace all of them to NewConstMetric() with type Counter? Am I understand right?

Most of them. Some should be gauges (but still with `NewConstMetric`) like time: https://www.robustperception.io/are-increasing-timestamps-counters-or-gauges That's why I said "most". Use your own judgment first, then we can go over corner cases.

DexterHD: Ack

AlekSi: There is also an example in mongodb_exporter itself: https://github.com/percona/mongodb_exporter/commit/27da9c342847f38cf13276f129b9578b6c6c5e9d

collector/common/op_counters.go
```diff
@@ -19,14 +19,14 @@ import (
 )
 
 var (
-	opCountersTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
+	opCountersTotal = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
AlekSi: > These numbers will grow over time

That's counter.

The same thing for one metric below. Please update it in a new PR.

CHANGELOG.md
```diff
@@ -7,10 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Changed
-- Moved `opcounters`, `opcountersRepl` to `common_collector` (#146), thx [@nikita-b](https://github.com/nikita-b). 
 - [PMM-3511](https://jira.percona.com/browse/PMM-3511): Switched to [mongo-go-driver](https://github.com/mongodb/mongo-go-driver).
-- [PMM-4168](https://jira.percona.com/browse/PMM-4168): Updated all dependencies including `github.com/prometheus/golang_client`. 
-As a result a lot of metrics was changed from Counter to ConstMetric. For more information look at commit [b8e8369](https://github.com/percona/mongodb_exporter/pull/145/commits/b8e836973e1e12608c24c0fe01437dfc49f88ae3).
```
AlekSi: The `Counter` is a metric type, that is important to users. `ConstMetric` is an implementation detail that is not.
Originally, that entry was about Counter => Gauge changes, but they were reverted.

PR #148, opened 2019-06-24
PMM-4236 Move test coverage to different stage.

DexterHD:
Here I have refactored `Makefile` and `.travis-ci.yml` a little bit, to improve CI speed.
You can compare builds [№740](https://travis-ci.com/percona/mongodb_exporter/builds/116636634) with [№733](https://travis-ci.com/percona/mongodb_exporter/builds/116619646). Look not only in total timing but also in different stages.

In the luckiest case, the total run time will be probably the same. About `7m30s`. 
But in the worst case (for example, if each Codeconv report will be working for `5m`) the new pipeline will be faster a lot because it runs only once, not after each testing job. Also, it would never run if some of the tests fail because it's in a different stage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ language: go
 stages:
   - build
   - test
-  - test-coverage
   - deploy
 
 go:
@@ -41,11 +40,6 @@ jobs:
       script:
         - make check-vendor-synced style build
 
-    - stage: test-coverage
-      script:
-        - make test-coverage
-      after_success: bash <(curl -s https://codecov.io/bash) -X fix
-
     - stage: deploy
       script: skip
       deploy:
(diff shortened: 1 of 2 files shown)


AlekSi:
Copying from https://jira.percona.com/browse/PMM-4236?focusedCommentId=239508&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-239508:

> Current PR runs tests with coverage information only for MongoDB 3.4. That means that we don't have code coverage information for version-specific code. That kind of defeats the purpose of testing with various MongoDB versions – we don't know if the version-specific code was tested or not.

DexterHD:
It seems it not so good idea, so I'm closing it.

PR #151, opened 2019-07-02
PMM-4131 Port features from dcu/mongodb_exporter.

DexterHD:
- [x] tcmalloc.go
- [x] conn_pool_stats.go
- [x] replset_conf.go
- [x] ~~collection_wiredtiger.go~~


collector/common/conn_pool_stats.go
```diff
@@ -0,0 +1,94 @@
...
+	numAScopedConnections.Set(stats.ASScopedConnections)
+	numAScopedConnections.Collect(ch)
+
+	totalInUse.Set(stats.TotalInUse)
+	totalInUse.Collect(ch)
+
+	totalAvailable.Set(stats.TotalAvailable)
+	totalAvailable.Collect(ch)
+
+	ch <- prometheus.MustNewConstMetric(totalCreatedDesc, prometheus.CounterValue, stats.TotalCreated)
```
AlekSi: Why this one metric is not like the others?
If it is because it is a counter, then constant counters can be created using the same method. As a bonus point that will not introduce new mutable package-level variables that you wanted to remove.

The same in other places.

collector/common/conn_pool_stats.go
```diff
@@ -0,0 +1,94 @@
...
+	totalAvailable.Describe(ch)
+	ch <- totalCreatedDesc
+}
+
+// GetConnPoolStats returns the server connPoolStats info.
+func GetConnPoolStats(client *mongo.Client) *ConnPoolStats {
+	result := &ConnPoolStats{}
+	err := client.Database("admin").RunCommand(context.TODO(), bson.D{{"connPoolStats", 1}, {"recordStats", 0}}).Decode(result)
+	if err != nil {
+		log.Error("Failed to get server status.")
```
AlekSi: Let's log `err`.

The same in other places.

collector/common/tcmalloc.go
```diff
@@ -0,0 +1,127 @@
...
+	tcmallocCacheBytes.WithLabelValues("thread_cache", "max_total").Set(m.Details.MaxTotalThreadCacheBytes)
+	tcmallocCacheBytes.WithLabelValues("thread_cache", "current_total").Set(m.Details.CurrentTotalThreadCacheBytes)
+	tcmallocCacheBytes.WithLabelValues("central_cache", "free").Set(m.Details.CentralCacheFreeBytes)
+	tcmallocCacheBytes.WithLabelValues("transfer_cache", "free").Set(m.Details.TransferCacheFreeBytes)
+	tcmallocCacheBytes.WithLabelValues("thread_cache", "free").Set(m.Details.ThreadCacheFreeBytes)
+	tcmallocCacheBytes.Collect(ch)
+
+	ch <- prometheus.MustNewConstMetric(tcmallocAggressiveDecommitDesc, prometheus.CounterValue, m.Details.AggressiveMemoryDecommit)
+	ch <- prometheus.MustNewConstMetric(tcmallocFreeBytesDesc, prometheus.CounterValue, m.Details.TotalFreeBytes)
+
```
AlekSi: Empty line.

collector/mongod/replset_conf.go
```diff
@@ -0,0 +1,126 @@
+package mongod
+
+import (
+	"context"
+
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+
```
AlekSi: Merge those groups.

collector/common/conn_pool_stats_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/percona/mongodb_exporter/testutils"
+)
+
+func TestGetConnPoolStatsDecodesFine(t *testing.T) {
+	// setup
+	t.Parallel()
```
AlekSi: That should be moved to subtests.

collector/common/conn_pool_stats_test.go
```diff
@@ -0,0 +1,52 @@
...
+	defer cancel()
+
+	t.Run("mongod", func(t *testing.T) {
+		// setup
+		defaultClient := testutils.MustGetConnectedMongodClient(ctx, t)
+		defer defaultClient.Disconnect(ctx)
+		// run
+		statusDefault := GetConnPoolStats(defaultClient)
+		// test
+		assert.NotNil(t, statusDefault)
```
AlekSi: Can we check at least _some_ values? This test (and others) will happily pass if an empty structure is returned.

collector/mongod/server_status.go
```diff
@@ -22,12 +22,12 @@ import (
...
 	"go.mongodb.org/mongo-driver/mongo"
 
-	collector_common "github.com/percona/mongodb_exporter/collector/common"
+	commoncollector "github.com/percona/mongodb_exporter/collector/common"
 )
 
 // ServerStatus keeps the data returned by the serverStatus() method.
 type ServerStatus struct {
-	collector_common.ServerStatus `bson:",inline"`
+	commoncollector.ServerStatus `bson:",inline"`
```
AlekSi: Why not `common` without named import?

mongodb_exporter.go
```diff
@@ -42,10 +42,11 @@ var (
...
 
-	collectDatabaseF   = kingpin.Flag("collect.database", "Enable collection of Database metrics").Bool()
-	collectCollectionF = kingpin.Flag("collect.collection", "Enable collection of Collection metrics").Bool()
-	collectTopF        = kingpin.Flag("collect.topmetrics", "Enable collection of table top metrics").Bool()
-	collectIndexUsageF = kingpin.Flag("collect.indexusage", "Enable collection of per index usage stats").Bool()
+	collectDatabaseF            = kingpin.Flag("collect.database", "Enable collection of Database metrics").Bool()
+	collectCollectionF          = kingpin.Flag("collect.collection", "Enable collection of Collection metrics").Bool()
+	collectTopF                 = kingpin.Flag("collect.topmetrics", "Enable collection of table top metrics").Bool()
+	collectIndexUsageF          = kingpin.Flag("collect.indexusage", "Enable collection of per index usage stats").Bool()
+	mongodbCollectConnPoolStats = kingpin.Flag("collect.connpoolstats", "Collect MongoDB connpoolstats").Bool()
```
AlekSi: Add `F` suffix as in other flags. Also make help text consistent with other `collect.XXX` flags.

collector/mongod/server_status.go
DexterHD: To add a specific sense. I decide that the word `common` by itself doesn't look good in this context.

collector/common/conn_pool_stats_test.go
DexterHD: Probably yes, but for this: 

1. We should force MongoDB to increase that numbers, as at least in the local environment all numbers returned by this command have `0` value.
2. We should change the test name.

collector/common/conn_pool_stats.go
DexterHD: > Why this one metric is not like the others?

Yes, it's because it is a counter.

> If it is because it is a counter, then constant counters can be created using the same method.

Sorry but I don't understand what exactly did you mean. Can you rephrase it? Should I change all other metrics by constant metrics or something else?

AlekSi:
```go
ch <- prometheus.MustNewConstMetric(syncClientConnectionsDesc, prometheus.GaugeValue, stats.SyncClientConnections)
```

Basically, https://jira.percona.com/browse/PMM-4167, but only for the new code in this PR.

collector/common/conn_pool_stats_test.go
AlekSi: Let's do that.

DexterHD: > We should force MongoDB to increase that numbers,

I've read the docs and to do that we should configure MongoDB in `replSet` or `cluster` mode, and connect together 2 or more servers. I think it some kind of overengineering for our testing environment at least in this case.

> This test (and others) will happily pass if an empty structure is returned.

To handle this case I've found than MongoDB [has a special field](https://docs.mongodb.com/manual/tutorial/use-database-commands/#command-responses) called `ok` for all responses that were succeeded. I've added this field in our tests. Now those tests wouldn't pass if an empty struct returned.

collector/mongos/database_status.go
```diff
@@ -109,15 +109,15 @@ func GetDatabaseStatList(client *mongo.Client) *DatabaseStatList {
...
+		log.Error(err)
 		return nil
 	}
 	for _, db := range dbNames {
 		dbStatus := DatabaseStatus{}
 		r := client.Database(db).RunCommand(context.TODO(), bson.D{{"dbStats", 1}, {"scale", 1}})
 		err := r.Decode(&dbStatus)
 		if err != nil {
-			log.Error("Failed to get database status.")
+			log.Error(err)
```
AlekSi: That will log something like "context deadline exceeded" without any context (pun intended). I would say we should use both message and error:
```go
log.Errorf("Failed to get database status: %s.", err)
```

collector/mongos/database_status.go
```diff
@@ -109,15 +109,15 @@ func GetDatabaseStatList(client *mongo.Client) *DatabaseStatList {
 	dbStatList := &DatabaseStatList{}
 	dbNames, err := client.ListDatabaseNames(context.TODO(), bson.M{})
 	if err != nil {
-		log.Error("Failed to get database names")
+		log.Error(err)
```
AlekSi: One more place.

collector/mongos/sharding_changelog.go
```diff
@@ -104,7 +104,7 @@ func GetShardingChangelogStatus(client *mongo.Client) *ShardingChangelogStats {
 
 	c, err := coll.Aggregate(context.TODO(), []bson.M{{"$match": match}, {"$group": group}})
 	if err != nil {
-		log.Error("Failed to execute find query on 'config.changelog'!")
+		log.Error(err)
```
AlekSi: And there.

AlekSi:
@BupycHuk PTAL

collector/common/conn_pool_stats_test.go
BupycHuk: But why they all have 0 value?

DexterHD: Because you need to have more than 1 server. Those metrics related to `replSet` or `cluster` mode, so to have connection pool you should have configured cluster first. 😄

collector/mongos/server_status_test.go
```diff
@@ -77,3 +81,17 @@ func loadServerStatusFromBson(data []byte, status *ServerStatus) {
 		panic(err)
 	}
 }
+
+func TestGetServerStatusDecodesFine(t *testing.T) {
+	// setup
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
BupycHuk: Why do we need this timeout?

DexterHD: 1) It's [standard approach](https://github.com/mongodb/mongo-go-driver#usage) to working with the new mongodb_driver.
2) To force disconnecting from MongoDB if something goes wrong. Without it, there is a potential risk that tests can become stuck.

PR #160, opened 2019-09-17
PMM-4702: Update mongo-driver to 1.1.1

DexterHD:


AlekSi:
https://github.com/Percona-Lab/pmm-submodules/pull/472. If it works, we will release 0.10 tomorrow morning

DexterHD:
> So, we will provide tls options via DSN, right?

I'm not sure how it should be. The "right way" from `mongodb` perspective is to use `connString`. Maybe we need to duplicate flags or map it somehow, but I don't know.

CHANGELOG.md
```diff
@@ -12,6 +12,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+## [0.10.0]
+### Changed
+- `go.mongodb.org/mongo-driver` was updated to `v1.1.1`.
+- All `--mongodb.tls*` flags was removed. Use [tls-options](https://docs.mongodb.com/manual/reference/connection-string/#tls-options) instead.
```
AlekSi: were

collector/mongodb_collector.go
```diff
@@ -53,16 +48,11 @@ type MongodbCollectorOpts struct {
...
-		TLSPrivateKeyFile:     in.TLSPrivateKeyFile,
-		TLSCaFile:             in.TLSCaFile,
-		TLSHostnameValidation: in.TLSHostnameValidation,
-		PoolLimit:             in.DBPoolLimit,
-		SocketTimeout:         in.SocketTimeout,
-		SyncTimeout:           in.SyncTimeout,
-		AuthentificationDB:    in.AuthentificationDB,
+		URI:                in.URI,
+		PoolLimit:          in.DBPoolLimit,
+		SocketTimeout:      in.SocketTimeout,
```
AlekSi: Those things are parts of URI too. If drivers supports then, we should remove them in 0.11 (later).

AlekSi:
Please make a release and a new branch for PMM ASAP.

collector/mongodb_collector.go
DexterHD: Ok. I've made a ticket.

PR #164, opened 2019-10-18
add lost connections metric and remove useless file

nikita-b:
Add active connections metrics and remove connections.go file from mongos module (it's same as connections.go in common module).


nikita-b:
@BupycHuk @AlekSi Hi! Could you review this small PR? :)

PR #193, opened 2020-07-13
PMM-6213 Remove go 1.13 from travis.

askomorokhov:
https://jira.percona.com/browse/PMM-6213


askomorokhov:
> Can we add master as second go version?
Yes, added.

PR #198, opened 2020-08-27
mongodb_mongos_sharding_chunks_total tweak

akira-kurogane:
This change is a proposal for the PMM-6361 issue. It adds a sort stage to the group-by aggregation behind the mongodb_mongos_sharding_chunks_total. The sort stage shouldn't be required logically, but it helps workaround a missing aggregation pipeline optimization.

Testing required:
1. Confirm the same query plan and execution time when it is run on standard mongodb clusters that don't have an index {shard: 1} on config.chunks.
2. Confirm there is a significant improvement (hopefully 10x at least) over the default behaviour (collscan) if used on a chunks collection of say 1 million docs when a {shard: 1} index is added

### DIFF
--- a/collector/mongos/sharding_topology.go
+++ b/collector/mongos/sharding_topology.go
@@ -128,7 +128,10 @@ func GetTotalChunks(client *mongo.Client) float64 {
 // GetTotalChunksByShard gets total chunks by shard.
 func GetTotalChunksByShard(client *mongo.Client) *[]ShardingTopoChunkInfo {
 	var results []ShardingTopoChunkInfo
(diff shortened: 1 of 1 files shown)


akira-kurogane:
PMM team - no need to act on this patch until Lalit C. and I finish proving if there is the expected speed benefit, or not.

akira-kurogane:
Testing has shown that a count by aggregation pipeline just isn't going to be fast (see Aug 28th comments in https://jira.percona.com/browse/PMM-6361).

I'll close this PR.

PR #201, opened 2020-09-02
PMM-6506 More metrics renaming

percona-csalguero:


exporter/metrics.go
```diff
@@ -271,6 +271,10 @@ func makeMetrics(prefix string, m bson.M, labels map[string]string, compatibleMo
 				continue
 			}
 
+			if renamedMetric := rawToCompatibleRawMetric(rm, specialConversions()); renamedMetric != nil {
```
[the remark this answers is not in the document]
BupycHuk: At least `rawToCompatibleRawMetric` should be renamed.

BupycHuk (changes requested):

exporter/metrics.go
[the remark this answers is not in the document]
percona-csalguero: Done

exporter/v1_compatibility.go
```diff
@@ -59,10 +54,10 @@ type conversion struct {
 	suffixMapping    map[string]string
 }
 
-func rawToCompatibleRawMetric(rm *rawMetric) *rawMetric {
+func metricRenameAndLabel(rm *rawMetric, convs []conversion) *rawMetric {
```
BupycHuk, line 57: it should be moved outside of this package since we use it not only for compatibility mode.

BupycHuk (changes requested):

exporter/v1_compatibility.go
percona-csalguero: Done. Moved to metrics.go

exporter/v1_compatibility.go
```diff
@@ -629,3 +588,65 @@ func walkTo(m primitive.M, path []string) interface{} {
 
 	return val
 }
+
+// specialConversions returns a list of special conversions we want to implement.
+// See: https://jira.percona.com/browse/PMM-6506
+func specialConversions() []conversion {
```
BupycHuk, line 594: let's move it too, since it's also not related to compatibility.

BupycHuk (changes requested):

BupycHuk (approved):

AlekSi (approved):

AlekSi:
I think we should do the same thing as in rds_exporter where we just dump all metrics in testdata and them compare them: https://github.com/percona/rds_exporter/tree/master/enhanced It will make PRs like that one much more clear – it will be visible in diff what metrics changed

PR #202, opened 2020-09-03
PMM-6608 Added more logs

percona-csalguero:
- Added error logs instead of stop collecting metrics in case of errors.
- Added options to disable getDiagnosticData and replicasetGetStatus collectors

### DIFF
--- a/exporter/diagnostic_data_collector.go
+++ b/exporter/diagnostic_data_collector.go
@@ -43,15 +43,14 @@ func (d *diagnosticDataCollector) Collect(ch chan<- prometheus.Metric) {
 	res := d.client.Database("admin").RunCommand(context.TODO(), cmd)
 
 	if err := res.Decode(&m); err != nil {
-		ch <- prometheus.NewInvalidMetric(prometheus.NewInvalidDesc(err), err)
+		d.logger.Errorf("cannot run getDiagnosticData: %s", err)
 		return
 	}
 
 	m, ok := m["data"].(bson.M)
 	if !ok {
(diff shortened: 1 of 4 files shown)


exporter/diagnostic_data_collector.go
```diff
@@ -43,15 +43,14 @@ func (d *diagnosticDataCollector) Collect(ch chan<- prometheus.Metric) {
 	res := d.client.Database("admin").RunCommand(context.TODO(), cmd)
 
 	if err := res.Decode(&m); err != nil {
-		ch <- prometheus.NewInvalidMetric(prometheus.NewInvalidDesc(err), err)
```
BupycHuk, line 46: Why can't we do both?
Log error and return invalid metric?

main.go
```diff
@@ -84,20 +87,26 @@ func main() {
 
 	if opts.URI == "" {
 		opts.URI = os.Getenv("MONGODB_URI")
```
BupycHuk, line 89: kong let's to override value by using `env` tag https://github.com/alecthomas/kong#supported-tags.

BupycHuk (changes requested):

exporter/diagnostic_data_collector.go
percona-csalguero: If we do that, prometheus will stop receiveing metrics so, we are preventing all other metrics to be exposed

BupycHuk: got it

BupycHuk (approved):

AlekSi (approved):
:shipit:

PR #207, opened 2020-09-09
PMM-6585 Always expose mongodb_up metric

percona-csalguero:
Created a new exporter named `generalCollector` where we can put any metric that's not related to any specific collector.
Also, this collector cannot be disabled so, the `mongodb_up` metric is going to be always exposed.
Fixed error from merging connection strategy and topology labels that made tests to fail. 

FB: https://github.com/Percona-Lab/pmm-submodules/pull/1097

PR #208, opened 2020-09-11
PMM-6585 Always expose mongodb_up metric

percona-csalguero:
New PR to replace https://github.com/percona/mongodb_exporter/pull/207
This is based on PMM-6610

### DIFF
--- a/exporter/general_collector.go
+++ b/exporter/general_collector.go
@@ -1,0 +1,58 @@
+// mongodb_exporter
+// Copyright (C) 2017 Percona LLC
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+package exporter
+
+import (
+	"context"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/readpref"
+)
+
+// This collector is always enabled and it is not directly related to any particular MongoDB
+// command to gather stats.
+type generalCollector struct {
+	ctx    context.Context
+	client *mongo.Client
+}
+
+func (d *generalCollector) Describe(ch chan<- *prometheus.Desc) {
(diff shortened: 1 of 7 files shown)


exporter/general_collector.go
```diff
@@ -0,0 +1,58 @@
...
+	var value float64
+
+	if err := client.Ping(ctx, readpref.PrimaryPreferred()); err == nil {
+		value = 1
+	}
+
+	d := prometheus.NewDesc("mongodb_up", "Whether MongoDB is up.", nil, nil)
+	up, err := prometheus.NewConstMetric(d, prometheus.GaugeValue, value)
+	if err != nil {
+		panic(err)
```
BupycHuk, line 52: We should not panic

BupycHuk (changes requested):

exporter/general_collector.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func (d *generalCollector) Collect(ch chan<- prometheus.Metric) {
+	ch <- mongodbUpMetric(d.ctx, d.client)
+}
+
+func mongodbUpMetric(ctx context.Context, client *mongo.Client) prometheus.Metric {
+	var value float64
+
+	if err := client.Ping(ctx, readpref.PrimaryPreferred()); err == nil {
```
BupycHuk, line 45: let's log error.

BupycHuk (changes requested):

exporter/general_collector.go
BupycHuk: @percona-csalguero it's not addressed yet.

percona-csalguero: done

BupycHuk: you are logging nil err

percona-csalguero: sorry .

exporter/general_collector.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func mongodbUpMetric(ctx context.Context, client *mongo.Client, log *logrus.Logger) prometheus.Metric {
+	var value float64
+
+	err := client.Ping(ctx, readpref.PrimaryPreferred())
+	if err == nil {
+		value = 1
+	}
+	log.Errorf("error while checking mongodb connection: %s", err)
```
BupycHuk, line 51: now you are logging even if there no error.

BupycHuk (approved):

AlekSi (changes requested):

exporter/general_collector.go
AlekSi: @BupycHuk @percona-csalguero It is totally fine to panic on programmer's mistakes (as opposed to environment) to make them visible. That's why we use `MustRegister`, and that's why prometheus package provides `MustNewConstMetric` that should be used there.

AlekSi (approved):
LGTM, please merge once CI passes.

But please also use `MustNewConstMetric` where it does not depend on environment (if MongoDB down, if version is incorrect, etc).

exporter/general_collector.go
percona-csalguero: :+1:

PR #210, opened 2020-09-15
Pmm 6643 high cpu usage 6610

percona-csalguero:


main.go
```diff
@@ -110,5 +115,23 @@ func main() {
 		log.Fatal(err)
 	}
 
+	if opts.ProfilerOutput != "" {
+		f, err := os.Create(opts.ProfilerOutput)
+		if err != nil {
+			log.Fatal(err)
+		}
+		if err := pprof.StartCPUProfile(f); err != nil {
```
AlekSi, line 123: We should use net/http/pprof instead. I think there is a code in exporter_shared for it; if not, please add it to that shared package

percona-csalguero: :+1:

PR #211, opened 2020-09-15
PMM-6643 High CPU usage

percona-csalguero:
Improve CPU usage


exporter/diagnostic_data_collector_test.go
```diff
@@ -68,8 +67,8 @@ mongodb_oplog_stats_wt_transaction_update_conflicts 0` + "\n")
 		"mongodb_oplog_stats_wt_btree_fixed_record_size",
 		"mongodb_oplog_stats_wt_transaction_update_conflicts",
 	}
-	// TODO: use NewPedanticRegistry when mongodb_exporter code fulfils its requirements (https://jira.percona.com/browse/PMM-6630).
```
BupycHuk, line 71: Why did you replace it by PedanticRegistry? Did you make it stable?

exporter/exporter_test.go
```diff
@@ -94,7 +94,7 @@ func TestConnect(t *testing.T) {
 			go func() {
 				defer wg.Done()
 				res, err := http.Get(ts.URL) //nolint:noctx
-				assert.Nil(t, e.client)
+				// assert.Nil(t, e.client) // TODO: Fix. Make it nil if not global connection pool
```
BupycHuk, line 97: please revert this change

exporter/metrics.go
```diff
@@ -353,64 +355,87 @@ func metricRenameAndLabel(rm *rawMetric, convs []conversion) *rawMetric {
...
-			oldName:     "mongodb_ss_opLatencies_ops",
-			prefix:      "mongodb_ss_opLatencies",
-			suffixLabel: "op_type",
-			suffixMapping: map[string]string{
-				"commands_ops":     "commands",
-				"reads_ops":        "reads",
-				"transactions_ops": "transactions",
-				"writes_ops":       "writes",
-			},
+func specialMetricRenameAndLabel(rm *rawMetric) *rawMetric {
```
BupycHuk, line 358: what's the difference of this method from MetricRename and Label?

exporter/utils_test.go
```diff
@@ -29,18 +33,11 @@ func collect(c helpers.Collector) []prometheus.Metric {
 	return m
 }
 
-func filterMetrics(metrics []*helpers.Metric, filters []string) []*helpers.Metric {
+func filterMetrics(metrics []*helpers.Metric) []*helpers.Metric {
 	res := make([]*helpers.Metric, 0, len(metrics))
 
 	for _, m := range metrics {
 		m.Value = 0
```
BupycHuk, line 40: Why did you set value of metric to 0?

exporter/utils_test.go
```diff
@@ -29,18 +33,11 @@ func collect(c helpers.Collector) []prometheus.Metric {
 	return m
 }
 
-func filterMetrics(metrics []*helpers.Metric, filters []string) []*helpers.Metric {
+func filterMetrics(metrics []*helpers.Metric) []*helpers.Metric {
```
BupycHuk, line 36: Any reason to still have this function?

BupycHuk (changes requested):

exporter/diagnostic_data_collector_test.go
percona-csalguero: Metric names are stable. We don't have any duplicated name nor problems with help.

exporter/metrics.go
percona-csalguero: They use a different set of rules. 
Since now rules are global to avoid passing/copying maps and the garbage collector work, I had to split into 2 different functions.

exporter/utils_test.go
percona-csalguero: to be able to compare all metric names

exporter/utils_test.go
percona-csalguero: I am going to rename it to zeroMetrics.
The idea is to collect all metrics, zero then and return a new list we can use to compare against a predefined list to ensure there are no missing metrics in the output. 
It will be useful when we improve the metrics renaming code.

exporter/diagnostic_data_collector_test.go
BupycHuk: Actually it also checks if we have same metrics in describe and collect, but we don't.

exporter/diagnostic_data_collector_test.go
```diff
@@ -68,54 +69,47 @@ mongodb_oplog_stats_wt_transaction_update_conflicts 0` + "\n")
...
+
+	reg := prometheus.NewPedanticRegistry()
 	err := reg.Register(c)
 	require.NoError(t, err)
 	err = testutil.GatherAndCompare(reg, expected, filter...)
 	assert.NoError(t, err)
 }
 
 func TestAllDiagnosticDataCollectorMetrics(t *testing.T) {
+	// The metrics list will  change depending on the sanbox.
```
BupycHuk, line 81:
```suggestion
	// The metrics list will  change depending on the sandbox.
```

exporter/exporter.go
```diff
@@ -157,7 +161,18 @@ func (e *Exporter) handler() http.Handler {
 		ctx := r.Context()
 
 		client := e.client
-		topologyInfo := e.topologyInfo
+
+		if err := e.loadTopologyInfo(ctx, client); err != nil {
```
BupycHuk, line 165: @percona-csalguero as far as I know we provide `global-conn-pool` flag to mongodb_exporter and it shouldn't affect anything. And based on discussion with @AlekSi if `global-conn-pool` is false then each request should be like separate world.

exporter/exporter.go
```diff
@@ -230,3 +233,58 @@ func connect(ctx context.Context, dsn string) (*mongo.Client, error) {
...
+			}
+		}()
+	}
+
+	wasRestarted, err := e.wasRestarted(ctx, client)
+	if err != nil {
+		return errors.Wrap(err, "cannot check if the MongoDB server was restarted")
+	}
+
+	if e.topologyInfo != nil && !wasRestarted {
```
BupycHuk, line 257: not thread-safe

exporter/exporter_test.go
BupycHuk: ^

exporter/metrics.go
```diff
@@ -265,24 +266,44 @@ func makeMetrics(prefix string, m bson.M, labels map[string]string, compatibleMo
 				continue
 			}
 
-			metrics := []*rawMetric{rm}
+			// metrics := []*rawMetric{rm}
```
BupycHuk, line 269: these changes were done to be able to keep `mongodb_mongod_instance_uptime_seconds` and `mongodb_instance_uptime_seconds` together.

exporter/utils_test.go
BupycHuk: ok

exporter/utils_test.go
```diff
@@ -29,32 +32,39 @@ func collect(c helpers.Collector) []prometheus.Metric {
...
-		}
+		res = append(res, m)
 	}
 
 	return res
 }
 
-func getMetricNames(lines []string) map[string]bool {
-	names := map[string]bool{}
+func readTestData(filename string, destination interface{}) error {
```
BupycHuk, line 49: can we make this function return `([]string, error)`?

BupycHuk (changes requested):

exporter/exporter.go
AlekSi: It would be the right thing to do, but it doesn't look like we can do it right now.

I tried to run `make test-cluster` and `MONGODB_URI=mongodb://127.0.0.1:17001 go run main.go --compatible-mode --mongodb.global-conn-pool` locally on macOS and measure time – `curl http://127.0.0.1:9216/metrics` takes 4 seconds (!!!). That's insane given that we want to scrape exporter every second.

I then took a profile (added pprof.samples.cpu.001.pb.gz to the ticket). It shows that most of the time is spent on starting and waiting for external processes from https://github.com/shirou/gopsutil. Unless it is something macOS specific, we should not start external processes on every scrape.

Please try it locally on Linux – `go tool pprof -http=: http://127.0.0.1:9216/debug/pprof/profile` doesn't work on Carlos' machine for some reason.

PR #214, opened 2020-09-18
PMM-6643 Removed call to Toolkit's GetHostInfo

percona-csalguero:


exporter/topology_info.go
```diff
@@ -31,6 +33,11 @@ const (
 	labelClusterID       = "cl_id"
 	labelReplicasetName  = "rs_nm"
 	labelReplicasetState = "rs_state"
+
+	typeIsDBGrid    = "isdbgrid"
+	typeMongos      = "mongos"
```
BupycHuk, line 38: can we replace them with enum?

main.go
```diff
@@ -110,5 +115,23 @@ func main() {
 		log.Fatal(err)
 	}
 
+	if opts.ProfilerOutput != "" {
```
AlekSi, line 118: I still think we should not expose this functionality this way. Let's remove it and figure out why net/http/pprof does not work for you

exporter/topology_info.go
percona-csalguero: Done

main.go
percona-csalguero: Done

exporter/topology_info.go
```diff
@@ -21,16 +21,25 @@ import (
...
 
+type mongoDBNodeType string
+
 const (
 	labelClusterRole     = "cl_role"
 	labelClusterID       = "cl_id"
 	labelReplicasetName  = "rs_nm"
 	labelReplicasetState = "rs_state"
+
+	typeIsDBGrid    mongoDBNodeType = "isdbgrid"
```
BupycHuk, line 39: imho, this still can be just a string since we don't return it as result.

exporter/diagnostic_data_collector.go
```diff
@@ -72,10 +71,14 @@ func (d *diagnosticDataCollector) Collect(ch chan<- prometheus.Metric) {
...
 		}
 
-		hi, err := util.GetHostInfo(d.ctx, d.client)
+		nodeType, err := getNodeType(d.ctx, d.client)
 		if err != nil {
-			d.logger.Errorf("cannot get host info: %s", err)
-		} else if hi.NodeType == util.TypeMongos {
+			d.logger.Errorf("Cannot get node type to check if this is a mongos: %s", err)
+
+			return
```
BupycHuk, line 78: this don't let us provide diagnostic data in case we are not able to get node type.

BupycHuk (approved):

AlekSi (approved):

PR #236, opened 2021-03-11
PMM-7284 connection string clarification

rnovikovP:


denisok:
#240 is there and pass testing. please close

PR #249, opened 2021-03-18
Fix issue with DirectConnect

masaruhoshi:
This is a fix for https://jira.percona.com/browse/PMM-7128

Ref.: https://pkg.go.dev/go.mongodb.org/mongo-driver/mongo/options#ClientOptions.SetDirect

TODO: When all checks have passed and code is ready for the review, please add `pmm-review-exporters` as the reviewer. That would assign people from the review team automatically.

### DIFF
--- a/exporter/exporter.go
+++ b/exporter/exporter.go
@@ -44,6 +44,7 @@ type Exporter struct {
 type Opts struct {
 	CompatibleMode          bool
 	GlobalConnPool          bool
+	DirectConnect           bool
 	URI                     string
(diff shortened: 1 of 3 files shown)


percona-csalguero (approved):

JiriCtvrtka (approved):

rnovikovP:
Looks like we added a new flag to the exporter. If this is a case, should we add it to https://github.com/percona/mongodb_exporter/blob/main/README.md as well?

PR #258, opened 2021-04-10
PMM-7278 X-509 authentication mechanism

percona-csalguero:
[PMM-7278](https://jira.percona.com/browse/PMM-7278) **mongo_exporter fails to authorize when Mongo running with authMechanism=MONGODB-X509**

- Added Authentication method to the command line parameters to enable using X509 certs.
- Added SkipTLSVerify and connection timeout for testing purposes.
- Updated dependencies.
- Updated Go version in the testing workflow. 
- Updated golangci-lint config due to a [deprecation notice in the project](https://golangci-lint.run/usage/configuration/): 
```
-linters:
# please, do not use `enable-all`: it's deprecated and will be removed soon.
# inverted configuration with `enable-all` and `disable` is not scalable during updates of golangci-lint
```
---
- [ ] Tests passed.


.github/workflows/go.yml
```diff
@@ -20,8 +20,8 @@ jobs:
       fail-fast: false
       matrix:
         go-version:
-          - 1.14.x
-          - 1.15.x
+          - 1.16.x
```
BupycHuk, line 23: both 1.16.x?

main.go
```diff
@@ -41,6 +42,8 @@ type GlobalFlags struct {
 	URI                   string `name:"mongodb.uri" help:"MongoDB connection URI" env:"MONGODB_URI" placeholder:"mongodb://user:pass@127.0.0.1:27017/admin?ssl=true"`
 	GlobalConnPool        bool   `name:"mongodb.global-conn-pool" help:"Use global connection pool instead of creating new pool for each http request."`
 	DirectConnect         bool   `name:"mongodb.direct-connect" help:"Whether or not a direct connect should be made. Direct connections are not valid if multiple hosts are specified or an SRV URI is used." default:"false"`
+	AuthMechanism         string `name:"mongodb.auth-mechanism" help:"Authentication mechanism. Example: MONGODB-X509."`
```
BupycHuk, line 45: Why don't pass Authmechanism as part of DSN?

BupycHuk: As far as I see mongo-driver knows how to parse it.

main.go
```diff
@@ -41,6 +42,8 @@ type GlobalFlags struct {
...
+	InsecureSkipVerify    bool   `name:"insercure-skip-verify" help:"Skip X509 certs" default:"true"`
```
BupycHuk, line 46: it's also can be done via URI. and why it's true by default?

BupycHuk (changes requested):

main.go
askomorokhov: AFAIK we should support two ways flag and URI: https://jira.percona.com/browse/PMM-7278?focusedCommentId=277074&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#/comment-277074

askomorokhov (approved):
Overall looks good to me. 👍 
Please fix a type with the go version.
And let's follow mongo client defaults:  `--insercure-skip-verify` is false if not specified.

JiriCtvrtka (approved):

percona-csalguero:
Closing because we are going to use the URI for all the parameters.

PR #259, opened 2021-04-15
More labels for replSetGetStatus() info

weastur:
Add common labels to members info parsed from `replSetGetStatus()`. Also, add `member_state` as a label if exists. Without these changes, members metrics almost useless because we have only `member_idx` label. So, for example, calculate replication lag as the difference between `optimeDate` fields isn't possible because we don't know who is primary and to which cluster this member belongs to.

### DIFF
--- a/exporter/metrics.go
+++ b/exporter/metrics.go
@@ -293,10 +293,10 @@ func makeMetrics(prefix string, m bson.M, labels map[string]string, compatibleMo
 
 // Extract maps from arrays. Only some structures like replicasets have arrays of members
 // and each member is represented by a map[string]interface{}.
-func processSlice(prefix, k string, v []interface{}, common_labels map[string]string, compatibleMode bool) []prometheus.Metric {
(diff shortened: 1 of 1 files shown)


percona-csalguero:
Hi. Thanks for the contribution.
Could you change the `_` to snake case please? We would like to follow Go rules to make the code consistent and avoid the annoying linter messages.
Thanks

weastur:
> Hi. Thanks for the contribution.
> Could you change the `_` to snake case please? We would like to follow Go rules to make the code consistent and avoid the annoying linter messages.
> Thanks

Fixed

askomorokhov (approved):
Hi! Thanks for the contribution.
Looks good to me. 👍

percona-csalguero (approved):

Aleksa-del:
@weastur Hello! We would love to send you a gift from Percona for your contributions. Please, contact us at community-team@percona.com

Issue #313, opened 2021-07-23
Prometheus exporter continues to export after database has been dropped


msmerc:
Hi,

I've noticed that if we remove a mongo database the exporter will still retain all its metrics. I've not tested to see what happens if we drop a collection - but it's probably the same.

We find our dev server ends up with lots of dead metrics as a result. Might it be possible to enhance the exporter so that it forgets about collections after they are removed?

percona-csalguero:
I cannot reproduce the issue.
I've created a new [branch and a PR](https://github.com/percona/mongodb_exporter/pull/323) with a test case. Please fill free to change it to make it prove the issue. 

Thanks

percona-csalguero:
Closing due to no response. I am assuming this cannot be reproduced.

PR #323, opened 2021-08-05
Added test to prove the isue #313

percona-csalguero:


percona-csalguero:
Closing it since it cannot be reproduced

PR #345, opened 2021-09-22
PMM-7116 myState metric always has a value

percona-csalguero:
[PMM-7116](https://jira.percona.com/browse/PMM-7116) MongoDB ReplSet Summary dashboard: incorrect STARTUP state

Currently, when there is no connection, the exporter is not exposing the current/real server state.
This PR fixes that issue and exposes the UNKNOWN state whenever the state cannot be determined.


exporter/diagnostic_data_collector.go
```diff
@@ -47,15 +47,13 @@ func (d *diagnosticDataCollector) Collect(ch chan<- prometheus.Metric) {
 	if err := res.Decode(&m); err != nil {
 		d.logger.Errorf("cannot run getDiagnosticData: %s", err)
 
-		return
+		// return
```
JiriCtvrtka, line 50: Remove this before merge.

JiriCtvrtka (approved):

BupycHuk (approved):

PR #351, opened 2021-10-04
Add docker image link to the Readme

AsterYujano:
As stated in this Issue: https://github.com/percona/mongodb_exporter/issues/64

A quick Docker documentation is a good enhancement to MongoDB exporter.

Related project card: https://github.com/orgs/percona/projects/2#card-66138726

[PMM-1026](https://jira.percona.com/browse/PMM-1026) (optional, if ticket reported)

- [ ] Links to other linked pull requests (optional).
---
- [ ] Tests passed.
- [ ] Fix conflicts with target branch.
- [ ] Update jira ticket description if needed.
- [ ] Attach screenshots/console output to confirm new behavior to jira ticket, if applicable.

When all checks have passed and code is ready for the review, bot should add `pmm-review-exporters` team as the reviewer. That would assign people from the review team automatically. Report any issues on our [Forums](https://forums.percona.com) and [Discord](https://discord.gg/mQEyGPkNbR).

Issue #264, opened 2021-05-07
Integrate Go releaser with Github Actions.


askomorokhov:
Integrate Go releaser with Github Actions.

percona-csalguero:
Don't we have this already? [release.yml](https://github.com/percona/mongodb_exporter/blob/main/.github/workflows/release.yml) file

Issue #64, opened 2017-05-29
Provide official docker image on docker hub


heartwilltell:
It would be really userful

Exordian:
+1 enabling an automated build on dockerhub should just take a few minutes

rnovikovP:
We have  similar requests for all  exporters so you can  follow https://jira.percona.com/browse/PMM-1026 to track progress

TheQueenIsDead:
Super old issue, but if anyone is looking for an image, [Bitnami](https://bitnami.com/) has one available

GitHub: https://github.com/bitnami/bitnami-docker-mongodb-exporter
Docker Hub: https://hub.docker.com/r/bitnami/mongodb-exporter/

Could be worth linking to it in the README, though most avid Googlers will likely find it

AsterYujano:
Hello, I was looking for a docker image recently (to provide Grafana insights of a MongoDB sharding) and indeed it would be nice to have it in the Readme, I am going to add it.

denisok:
percona/mongodb_exporter:0.20

nick4fake:
Is there any guide? Looks like there is no instruction on Docker Hub

denisok:
@nick4fake standard README have some instructions: https://github.com/percona/mongodb_exporter#docker

PR #354, opened 2021-10-04
build container image on release

kbudde:
-  #264
- #64

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,6 +21,12 @@ jobs:
         uses: percona-platform/setup-go@v2
         with:
           go-version: 1.16
+      - 
+        name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -21,6 +21,12 @@ jobs:
         uses: percona-platform/setup-go@v2
         with:
           go-version: 1.16
+      - 
+        name: Login to Docker Hub
+        uses: docker/login-action@v1
```
denisok, line 26:
```suggestion
        uses: percona-platform/login-action@v1
```

we fork it for the security reasons.

.github/workflows/release.yml
```diff
@@ -21,6 +21,12 @@ jobs:
...
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
```
denisok, line 28: ok, so I would need to create those secrets.

denisok:
also closes #351

.github/workflows/release.yml
denisok: opened IT ticket to get creds

PR #366, opened 2021-10-15
prevent exporter exit after start

### LINKED ISSUE #361: mongodb_exporter fails to start if mongo is down during startup period
at startup, if mongo is not listening yet, mongodb_exporter fails to start:

```2021/10/12 01:43:10 server selection error: server selection timeout, current topology: { Type: Single, Servers: [{ Addr: full.server.fqdn:27017, Type: Unknown, Average RTT: 0, Last error: connection() error occured during connection handshake: dial tcp 127.123.123.123:27017: connect: connection refused }, ] }
```

In my experience, the point of exporters is to connect to the existing monitored instance process at the connect-time of a scraper, and relay the relevant metrics at that time.  Failure absent a metrics request seems like the cart is being put before the horse, here.  This also breaks the eventually-consistent model.. Using this, I have to gate the startup of the exporter until after my mongodb instance..  that doesn't make any sense whatsoever.

kbudde:
- Fixes #361: with `--mongodb.global-conn-pool` option set, the exporter only tried to connect once after start
- topology now refreshes for every scrape as it might change (same behaviour as without the option `--mongodb.global-conn-pool`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,13 +37,13 @@ Currently, these metric sources are implemented:
 |\-\-mongodb.global-conn-pool|Use global connection pool instead of creating new connection for each http request.||
 |\-\-web.listen-address|Address to listen on for web interface and telemetry|\-\-web.listen-address=":9216"|
 |\-\-web.telemetry-path|Metrics expose path|\-\-web.telemetry-path="/metrics"|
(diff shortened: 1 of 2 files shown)


in4mer:
May we please have this patched into a stable like 0.20, as it's addressing a bug there?

PR #378, opened 2021-11-11
PMM-7802 arbiter authentication

percona-csalguero:
[PMM-7802: mongodb_exporter v0.20+ cannot start on MongoDB arbiter nodes](https://jira.percona.com/browse/PMM-7802)

- [x] Tests passed.
- [x] Feature build: https://github.com/Percona-Lab/pmm-submodules/pull/2139


exporter/topology_info_test.go
```diff
@@ -29,19 +29,75 @@ import (
...
+				labelReplicasetState: "1",
+				labelClusterRole:     "configsvr",
+				labelClusterID:       "f",
+			},
+		},
+		{
+			containerName: "mongo-1-arbiter",
+			want: map[string]string{
+				labelReplicasetName:  "",
+				labelReplicasetState: "0",
```
ShashankSinha252, line 67: As per docs, this should be [7](https://docs.mongodb.com/v4.4/reference/replica-states/). Correct me if I'm wrong.

exporter/diagnostic_data_collector.go
```diff
@@ -43,6 +43,11 @@ func (d *diagnosticDataCollector) Collect(ch chan<- prometheus.Metric) {
 
 	cmd := bson.D{{Key: "getDiagnosticData", Value: "1"}}
 	res := d.client.Database("admin").RunCommand(d.ctx, cmd)
+	if res.Err() != nil {
+		if isArbiter, _ := isArbiter(d.ctx, d.client); isArbiter {
+			return
```
ShashankSinha252, line 48: How about we log that we hit an arbiter?

ShashankSinha252 (changes requested):

exporter/diagnostic_data_collector.go
percona-csalguero: No. This runs every 1 minute and we would fill the logs.

ShashankSinha252: Got it.

exporter/topology_info_test.go
percona-csalguero: No, because we are getting it from `getDIagnosticData` but `getDiagnosticData` doesn't work on arbiters so we are getting a 0

ShashankSinha252 (commented):
Need change to remove the extra line. LGTM otherwise.

JiriCtvrtka (approved):

ShashankSinha252 (approved):

exporter/topology_info.go
```diff
@@ -110,9 +110,16 @@ func (t *topologyInfo) loadLabels(ctx context.Context) error {
 		t.labels[labelReplicasetName] = rs.Config.ID
 	}
 
+	isArbiter, err := isArbiter(ctx, t.client)
+	if err != nil {
+		return err
+	}
+
 	cid, err := util.ClusterID(ctx, t.client)
 	if err != nil {
```
BupycHuk, line 119:
```suggestion
	if err != nil && !isisArbiter {
```

exporter/diagnostic_data_collector.go
BupycHuk: What about logging to Debug level?

Issue #380, opened 2021-11-12
Container: Passwort file as source of the mongodb passwort


Andy-Voigt:
Hello,

it would be relay helpful if the docker container can read a password file as source of the mongodb password.

denisok:
Hi @Andy-Voigt,
thanks for the idea. Could you please help me to understand what format that password file should be ? keyfile? some specific format? Or just what evere?
Thanks,
Denys

Andy-Voigt:
Hello @denisok, 

Docker mounts secrets (passwords) under /tmp/SECRET_NAME as plain text file. This is done to keep the envoriment variables free from passwords. If we cannot use this mechanism we have to write the mongodb passwort to our compose file.

johi12:
@Andy-Voigt One way you can solve this is to do it as it's done in the Grafana docker image, they have a shell script as an entrypoint and that script contains this, https://github.com/grafana/grafana/blob/f5641c0293e5c2218b1e2506e433df51390ff63f/packaging/docker/run.sh#L49-L63

Then you can pass the path to the secret as an environment variable.

Andy-Voigt:
Thanks for your replay. Thats pretty close to what we have done today. We "fork" the original Docker image and copy the mongodb_exporter to an ubuntu image. With that we create a bash script that fetch a config file from our spring config server and grep out the URL, password and username for mongodb. Now we can run that "sidecar" as normal service in our swarm. 

But it would be much simpler if we can use a docker secrete.

denisok:
OK, now I got it.
https://docs.docker.com/engine/swarm/secrets/
https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/

make sense now to add both env and read from file options to mongodb_exporter so it could pick up secrets and maybe also config from file or from the env vars.

denisok:
user and pass are in #560 . FILEs are still missed

gthieleb:
Are there any news on that? Would it be possible to address multiple db instances with using a single secret?

PR #392, opened 2021-12-07
Add build deb and rpm packages.

anton-bystrov:
[PMM-5598](https://jira.percona.com/browse/PMM-5598)

Add building deb and rpm packages.

---

- [ ] Tests passed.
- [x] Fix conflicts with target branch.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -40,3 +40,23 @@ dockers:
   - "percona/mongodb_exporter:{{.Version}}"
   dockerfile: Dockerfile
   use: buildx
+nfpms:
+  - file_name_template: "{{ .ProjectName }}-{{ .Version }}.{{ .Os }}-{{ .Arch }}"
+    homepage:  https://www.percona.com/
+    description: MongoDB exporter for Prometheus
+    maintainer: Percona
+    vendor: Percona
+    license: Apache 2.0
+    formats:
+    - deb
+    - rpm
+    bindir: /usr/bin
+    replacements:
+      amd64: 64-bit
+      darwin: macOS
+    scripts:
+      postinstall: ".scripts/postinst"
+    contents:
+      - src: .scripts/systemd/mongodb_exporter.service
+        dst: /etc/systemd/system/mongodb_exporter.service
+        type: "config|noreplace"
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -53,7 +53,6 @@ nfpms:
     bindir: /usr/bin
     replacements:
       amd64: 64-bit
-      darwin: macOS
     scripts:
       postinstall: ".scripts/postinst"
     contents:
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -60,3 +60,5 @@ nfpms:
       - src: .scripts/systemd/mongodb_exporter.service
         dst: /etc/systemd/system/mongodb_exporter.service
         type: "config|noreplace"
+      - src: .scripts/default/mongodb_exporter.example
+        dst: /etc/default/mongodb_exporter.example
--- a/.scripts/postinst
+++ b/.scripts/postinst
@@ -1,0 +1,35 @@
+#!/bin/sh
+
+set -e
+
+# summary of how this script can be called:
+#        * <postinst> `configure' <most-recently-configured-version>
+#        * <old-postinst> `abort-upgrade' <new version>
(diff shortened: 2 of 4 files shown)


.scripts/systemd/mongodb_exporter.service
```diff
@@ -0,0 +1,20 @@
...
+After=network.target
+
+[Service]
+Type=simple
+
+User=pmm
+Group=pmm
+
+ExecStart=/usr/bin/mongodb_exporter \
+  --mongodb.uri=mongodb://user:pass@127.0.0.1:27017/admin?ssl=true \
```
denisok, line 13: interesting... this probably should come from the env variables or conf file.

.goreleaser.yml
```diff
@@ -40,3 +40,23 @@ dockers:
...
+    maintainer: Percona
+    vendor: Percona
+    license: Apache 2.0
+    formats:
+    - deb
+    - rpm
+    bindir: /usr/bin
+    replacements:
+      amd64: 64-bit
+      darwin: macOS
```
denisok, line 56: is this needed for deb and rpm?

.scripts/postinst
```diff
@@ -0,0 +1,35 @@
...
+#     installation fails and the `postinst' is called with `abort-upgrade',
+#     `abort-remove' or `abort-deconfigure'.
+
+echo "Creating user and group..."
+
+GROUPNAME="pmm"
+USERNAME="pmm"
+
+getent group "$GROUPNAME" >/dev/null || groupadd -r "$GROUPNAME"
+getent passwd "$USERNAME" >/dev/null || useradd -r -g "$GROUPNAME" -M -s /sbin/nologin -c "pmm service user" "$USERNAME"
```
nikita-b, line 29: It looks overcomplicated. You can use single command:
``` 
   adduser --quiet --no-create-home  --group -c"PMM User" pmm
```

nikita-b:
What's about the defaults file?

.scripts/postinst
```diff
@@ -0,0 +1,35 @@
...
+#          <conflicting-package> <version>
+# for details, see http://www.debian.org/doc/debian-policy/ or
+# the debian-policy package
+#
+# quoting from the policy:
+#     Any necessary prompting should almost always be confined to the
+#     post-installation script, and should be protected with a conditional
+#     so that unnecessary prompting doesn't happen if a package's
+#     installation fails and the `postinst' is called with `abort-upgrade',
+#     `abort-remove' or `abort-deconfigure'.
```
nikita-b, line 21: Do we need this comment?

.scripts/postinst
```diff
@@ -0,0 +1,35 @@
...
+
+echo "Creating user and group..."
+
+GROUPNAME="pmm"
+USERNAME="pmm"
+
+getent group "$GROUPNAME" >/dev/null || groupadd -r "$GROUPNAME"
+getent passwd "$USERNAME" >/dev/null || useradd -r -g "$GROUPNAME" -M -s /sbin/nologin -c "pmm service user" "$USERNAME"
+
+systemctl daemon-reload > dev/null
```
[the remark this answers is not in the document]
nikita-b: Debian has a lot of helper scripts. For example:
http://manpages.ubuntu.com/manpages/bionic/man1/deb-systemd-invoke.1p.html

The trick is that we need RPM and DEB packages together...

FPM can create systemd service itself but it seems that goreleaser NFPM can't...

nikita-b:
A couple of points:
1. We need to register a systemd service.
2. We need to support upgrading to a new version (even for systemd service)
3.  We need to give the user the ability to overwrite the default settings (default file)

nikita-b (changes requested):

denisok:
> A couple of points:
> 
>     1. We need to register a systemd service.
> 
>     2. We need to support upgrading to a new version (even for systemd service)
> 
>     3. We need to give the user the ability to overwrite the default settings (default file)

maybe we are over complicating it.  Do we really need service file on thi stage? mongo exporter itself doesn't support environment or passing config file or parameter as a file (see #380).

I would probably remove service and get service later when mongodb_exporter is better parametrized, like taking user, url, pass from env.

.goreleaser.yml
anton-bystrov: I think this naming is more beautiful...no more.

denisok (approved):

nikita-b (approved):

Issue #319, opened 2021-07-28
use prometheus exporter-toolkit lib instead of exporter_shared


denisok:
To get closer to the Prometheus community we could switch from custom library to new community lib.
Also that would fix this issue:
https://jira.percona.com/browse/PMM-7284

PR #393, opened 2021-12-13
serve metrics handler with prometheus exporter-toolkit

rlankfo:
This PR serves the metrics handler utilizing `prometheus/exporter-toolkit` instead of `percona/exporter_shared`. This is a start towards fixing https://github.com/percona/mongodb_exporter/issues/319, at least partially, however this does not remove all references to `exporter_shared`. 

Note that TLS Config still needs to be handled by adding an option to pass the tlsConfigPath.

- [ ] Links to other linked pull requests (optional).

---

- [ ] Tests passed.
- [ ] Fix conflicts with target branch.

Once all checks pass and the code is ready for review, please add `pmm-review-exporters` team as the reviewer. That would assign people from the review team automatically. Report any issues on our [Forum](https://forums.percona.com) or [Discord](https://per.co.na/discord).


percona-csalguero:
Hi, 
Can you remove `go-kit/log` dependency. Since it is used only once, I think it is better to keep 3rd party dependencies at minimum.

Thanks.

exporter/exporter.go
```diff
@@ -21,11 +21,14 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"os"
 	"sync"
 
-	"github.com/percona/exporter_shared"
+	"github.com/go-kit/log/level"
```
percona-csalguero, line 27: Can we remove this dependency?
It is used only once so, I think it is better to keep 3rd party dependencies at minimum.

Thanks.

percona-csalguero (changes requested):

percona-csalguero:
I really appreciate this PR. Using exporter shared was preventing us to implement some other features like SSL support.
Thanks @rlankfo

exporter/exporter.go
rlankfo: updated :+1:

rlankfo: Note this will still get pulled in as a transitive dependency of `prometheus/exporter-toolkit`

percona-csalguero (approved):

denisok:
@JiriCtvrtka please review

.gitignore
```diff
@@ -6,3 +6,4 @@ bin
 build
 dist
 mongodb_exporter
+vendor
```
JiriCtvrtka, line 9: This is not necessary since we dont have vendor folder anymore.

JiriCtvrtka (approved):
LGTM only one minor comment.

denisok:
@percona-csalguero would test it a little bit more and merge it

PR #396, opened 2021-12-21
PMM-9312 default collectors

percona-csalguero:
[PMM-XXXX](https://jira.percona.com/browse/PMM-XXXX) (optional, if ticket reported)

- [ ] Links to other linked pull requests (optional).

---

- [ ] Tests passed.
- [ ] Fix conflicts with target branch.

Once all checks pass and the code is ready for review, please add `pmm-review-exporters` team as the reviewer. That would assign people from the review team automatically. Report any issues on our [Forum](https://forums.percona.com) or [Discord](https://per.co.na/discord).


exporter/common.go
```diff
@@ -13,62 +14,106 @@ import (
...
+// namespaces list. Namespaces have the form of <db>.<collection> and the collection name
+// can have a dot. Example: db1.collection.one -> db: db1, collection: collection.one
+// db1, db2.col2, db3.col.one will produce [db1, db2, db3]
+//
+func makeDBsFilter(filterInNamespaces []string) *primitive.E {
+	dbs := []string{}
+
+	for _, namespace := range filterInNamespaces {
+		parts := strings.Split(namespace, ".")
+		if len(parts) > 0 {
```
JiriCtvrtka, line 95: Is this if necessary?

JiriCtvrtka (approved):
LGTM just check my comment and lint comments.

exporter/common.go
```diff
@@ -13,62 +14,106 @@ import (
...
+
+	dbNames, err := client.ListDatabaseNames(ctx, filter, opts)
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot get the database names list")
+	}
+
+	return dbNames, nil
+}
+
+func makeExcludeFilter(exclude []string) *primitive.E {
```
BupycHuk, line 70: can we use `$nin` instead of `$ne`?

exporter/common.go
```diff
@@ -13,62 +14,106 @@ import (
...
-	return dbNames, nil
+	if len(dbs) == 0 {
+		return nil
+	}
+
+	return &primitive.E{Key: "name", Value: primitive.E{Key: "$in", Value: dbs}}
 }
 
-func listAllCollections(ctx context.Context, client *mongo.Client, filter []string) (map[string][]string, error) {
+func listAllCollections(ctx context.Context, client *mongo.Client, filterInNamespaces []string, excludeDBs []string) (map[string][]string, error) {
```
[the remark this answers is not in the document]
BupycHuk: looks like this comment is correct

exporter/exporter.go
```diff
@@ -117,26 +125,41 @@ func (e *Exporter) makeRegistry(ctx context.Context, client *mongo.Client, topol
...
+		}
+		e.opts.EnableDiagnosticData = true
+		e.opts.EnableDBStats = true
+		e.opts.CollectorTopMetrics = true
+		e.opts.EnableReplicasetStatus = true
 	}
 
 	// if we manually set the collection names we want or auto discovery is set
-	if (len(e.opts.CollStatsCollections) > 0 || e.opts.DiscoveringMode) && enableCollStats {
+	if (len(e.opts.CollStatsNamespaces) > 0 || e.opts.DiscoveringMode) && e.opts.EnableCollStats && limitsOk {
```
BupycHuk, line 158: it looks weird that when user just adds `--collector.collstats` without limits it doesn't enable it. 
I think in exporter if user doesn't provide limit and calls --collector.collstats` it should be unlimited.
The same for `collector.indexstats.`
@rnovikovP @denisok your thoughts?

exporter/indexstats_collector_test.go
```diff
@@ -44,7 +44,9 @@ func TestIndexStatsCollector(t *testing.T) {
 	ti := labelsGetterMock{}
 
 	database := client.Database("testdb")
-	database.Drop(ctx) //nolint
+	database.Drop(ctx)       //nolint:errcheck
+	defer database.Drop(ctx) //nolint:errcheck
```
BupycHuk, line 48: let's add a comment that database will be created automatically.

main.go
```diff
@@ -35,22 +35,29 @@ var (
...
+	WebListenAddress      string `name:"web.listen-address" help:"Address to listen on for web interface and telemetry" default:":9216"`
+	WebTelemetryPath      string `name:"web.telemetry-path" help:"Metrics expose path" default:"/metrics"`
+	LogLevel              string `name:"log.level" help:"Only log messages with the given severuty or above. Valid levels: [debug, info, warn, error, fatal]" enum:"debug,info,warn,error,fatal" default:"error"`
+
+	EnableDiagnosticData   bool `name:"collector.diagnosticdata" help:"Disable collecting metrics from getDiagnosticData"`
+	EnableReplicasetStatus bool `name:"collector.replicasetstatus" help:"Disable collecting metrics from replSetGetStatus"`
+	EnableDBStats          bool `name:"collector.dbstats" help:"Enable collecting metrics from dbStats"`
+	EnableTopMetrics       bool `name:"collector.topmetrics" help:"Enable collecting metrics from top admin command"`
+	EnableIndexStats       bool `name:"collector.indexstats" help:"Enablle collecting metrics from $indexStats"`
```
BupycHuk, line 51:
```suggestion
	EnableIndexStats       bool `name:"collector.indexstats" help:"Enable collecting metrics from $indexStats"`
```

main.go
```diff
@@ -35,22 +35,29 @@ var (
...
+	GlobalConnPool        bool   `name:"mongodb.global-conn-pool" help:"Use global connection pool instead of creating new pool for each http request." negatable:""`
+	DirectConnect         bool   `name:"mongodb.direct-connect" help:"Whether or not a direct connect should be made. Direct connections are not valid if multiple hosts are specified or an SRV URI is used." default:"true" negatable:""`
+	WebListenAddress      string `name:"web.listen-address" help:"Address to listen on for web interface and telemetry" default:":9216"`
+	WebTelemetryPath      string `name:"web.telemetry-path" help:"Metrics expose path" default:"/metrics"`
+	LogLevel              string `name:"log.level" help:"Only log messages with the given severuty or above. Valid levels: [debug, info, warn, error, fatal]" enum:"debug,info,warn,error,fatal" default:"error"`
+
+	EnableDiagnosticData   bool `name:"collector.diagnosticdata" help:"Disable collecting metrics from getDiagnosticData"`
+	EnableReplicasetStatus bool `name:"collector.replicasetstatus" help:"Disable collecting metrics from replSetGetStatus"`
```
BupycHuk, line 48: looks like something wrong with help message

main.go
```diff
@@ -35,22 +35,29 @@ var (
...
+	EnableDBStats          bool `name:"collector.dbstats" help:"Enable collecting metrics from dbStats"`
+	EnableTopMetrics       bool `name:"collector.topmetrics" help:"Enable collecting metrics from top admin command"`
+	EnableIndexStats       bool `name:"collector.indexstats" help:"Enablle collecting metrics from $indexStats"`
+	EnableCollStats        bool `name:"collector.collstats" help:"Enable collecting metrics from $collStats"`
+
+	CollectAll bool `name:"collect-all" help:"Enable all collectors. Same as specifying all --collector.<name>"`
+
+	CollStatsLimit int `name:"collector.collstats-limit" help:"Enable collstats and indexstats collector only if there are less than <n> collections. -1=No limit" default:"0"`
```
BupycHuk, line 56: I would rephrase it like `disable collstats and indexstats collector if there are more than <n> collections.`

exporter/common.go
BupycHuk: +1, even if there is no dot len(parts) is always > 0

exporter/common.go
```diff
@@ -13,62 +14,106 @@ import (
...
+		parts := strings.Split(namespace, ".")
+		if len(parts) > 0 {
+			dbs = append(dbs, parts[0])
+		}
 	}
 
-	return dbNames, nil
+	if len(dbs) == 0 {
+		return nil
+	}
```
BupycHuk, line 102: can it be in the beginning of the function?

BupycHuk (changes requested):
we need more explanation in docs how does discovery mode works. How does enabling and disabling of collectors work in combinations.

main.go
denisok: default:0 disabled?

exporter/exporter.go
denisok: should we shift that logic to pmm? it should check limits. For exporter I think you are right, but chances that someone shoots himself in the leg are high.
But yeah, I would do the logic in pmm and have it set limit flags to disabled and leave it here enabled by default.

BupycHuk: I don't mean that we should enable collstats by default I mean that user should be able to enable it just by calling one flag instead of two.
How it works now
`mongodb_expoter --collector.collstats --collector.collstats-limit=-1`
How it should work
`mongodb_expoter --collector.collstats`

So if you don't pass collstats limit collstats collector won't be enabled.

main.go
BupycHuk: right now it is, but it makes no sense. Why does user might want to enable collstats and disable it by setting limit to 0 in the same time?

denisok (approved):

BupycHuk (approved):

PR #407, opened 2022-01-10
Pmm 9400 on top of 2.26

percona-csalguero:


denisok (changes requested):
I changed the base. But there is still commit: 
```
serve metrics handler with prometheus exporter-toolkit (#393) 
rlankfo and percona-csalguero committed 3 days ago
```
it probably shouldn't be there

PR #410, opened 2022-01-11
Revert "Revert "serve metrics handler with prometheus exporter-toolki…

percona-csalguero:
Re-apply PR #393 

PR-393 was merged before we can use it properly from pmm-managed.
In order to make it work with basic auth, we need to implement the
changes in pmm-managed/mongodb exporter config to create a new file
having basic auth credentials.

This reverts commit 5e5bf25c7cfe3946ef6935632ab3d649db2ecc85. to reapply the original patch.


denisok:
lets merge it anyway, in pmm we use `release` branches, we can revert it there. and we shoud just add support for it. @percona-csalguero could you please open new Jira ticket and describe what needs to be added to support this?

PR #413, opened 2022-01-12
PMM-5471 add license checks

idoqo:
[PMM-5471 Add license checks to mongodb_exporter](https://jira.percona.com/browse/PMM-5471)

Adds the license header text to all existing go files without one, and include the license checker in CI.

### DIFF
--- a/.github/check-license.go
+++ b/.github/check-license.go
@@ -33,10 +33,9 @@ import (
 var (
 	generatedHeader = regexp.MustCompile(`^// Code generated .* DO NOT EDIT\.`)
 
-	projectInfo = `// mongodb_exporter
+	copyrightText = `// mongodb_exporter
 // Copyright (C) 2022 Percona LLC
-`
-	copyrightText = `//
+//
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU Affero General Public License as published by
 // the Free Software Foundation, either version 3 of the License, or
@@ -50,22 +49,33 @@ var (
 // You should have received a copy of the GNU Affero General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 `
+
+	copyrightPattern = regexp.MustCompile(`^// mongodb_exporter
+// Copyright \(C\) \d{4} Percona LLC
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// \(at your option\) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+`)
 )
 
-func checkHeader(path string, header string) bool {
+func checkHeader(path string) bool {
 	f, err := os.Open(path)
 	if err != nil {
 		log.Fatal(err)
 	}
 	defer f.Close()
 
-	// skip the lines containing the project name and copyright year
-	_, err = f.Seek(int64(len(projectInfo)), io.SeekStart)
-	if err != nil {
-		return false
-	}
-
-	actual := make([]byte, len(header))
+	actual := make([]byte, len(copyrightText))
 	_, err = io.ReadFull(f, actual)
 	if err == io.ErrUnexpectedEOF {
 		err = nil // some files are shorter than license header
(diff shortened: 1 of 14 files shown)


artemgavrilov (approved):

exporter/common.go
```diff
@@ -1,3 +1,19 @@
+// mongodb_exporter
+// Copyright (C) 2017 Percona LLC
```
denisok, line 2: 2017 ? is that correct @atymchuk ?

JiriCtvrtka (approved):

percona-csalguero (approved):

exporter/common.go
atymchuk: Actually, most of our client-related code is licensed under Apache-2.0, unless it's a fork and the upstream repo has a different license, which we need to stick to. Otherwise, let's make it Apache-2.0.

denisok: yeah, I am not and expert, just asking why 2017 if we have 2022 :)

atymchuk: Oh, I was answering the wrong question )
It doesn't make too much of a difference from the legal standpoint, but it's more logical to put 2022 :)

atymchuk: You can take it from [here](https://github.com/percona/pmm-agent/blob/main/actions/actions.go#L1-L14) except for the year, which should be 2022.

denisok: yes, that is my point :) if we adding new one it should have recent year

denisok (changes requested):
please change the year to 2022

exporter/common.go
idoqo: > please change the year to 2022

interestingly, it checks that the license text in the target file is the same as the one in `check-license.go` (currently it's 2017 which is why CI failed) - We can keep all of them at 2017, update them all to 2022, or make `check-license.go` independent of the year.

denisok: for me it should generate current year and check for that, and fail if there are some old headers. We probably need this check license centralized somewhere or in form of GH package to use in the all projects to not update it everywhere.

denisok: @atymchuk what do you think ?

idoqo: I think if we want to centralize this, we'd have to add support for different licenses (since not all projects use the same license). 

Also, maybe we can keep the old headers as they are without failing them and only use newer copyright years in new files? I updated check-license to accept all years from the oldest file to $CURRENT_YEAR so you could help check if it makes sense

.github/check-license.go
```diff
@@ -81,11 +77,13 @@ func checkHeader(path string, header string) bool {
 		return true
 	}
 
-	if header != string(actual) {
-		log.Print(path)
-		return false
+	for year := startYear; year <= time.Now().Year(); year++ {
```
denisok, line 80: I would say that is over complicated. I am too optimistic but in 100 years it would be 104 times... And if this would have thousands of files.

I could suggest just cut the date from the header of the file and compare it with template that doesn't have a date. Or maybe some smarter thing - like returning only diff between the strings and if it is a valid year - it is cool. What ever, but it should be also fast and simple enough - no need to consume a lot of CPUs cycle.

denisok (changes requested):

.github/check-license.go
idoqo: I went with the first approach and it seemed more straightforward to me.

.github/check-license.go
```diff
@@ -67,6 +59,12 @@ func checkHeader(path string, header string) bool {
 	}
 	defer f.Close()
 
+	// skip the lines containing the project name and copyright year
+	_, err = f.Seek(int64(len(projectInfo)), io.SeekStart)
+	if err != nil {
+		return false
+	}
+
```
BupycHuk, line 67: can we use regexp or another way to check these two lines?

idoqo: Another way could be to read and discard the first two lines with `io.ReadFull` (though I don't think that's better than this).

With regexp, I think we could use the copyright text as a pattern and allow for any year (something like `\d{4}`), but I can't quite figure how to prepare a multiline regex that `regexp.MustCompile` accepts yet

.github/check-license.go
```diff
@@ -21,53 +21,61 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+`
 
-var generatedHeader = regexp.MustCompile(`^// Code generated .* DO NOT EDIT\.`)
+	copyrightPattern = regexp.MustCompile(`^// mongodb_exporter
+// Copyright \(C\) \d{4} Percona LLC
```
BupycHuk, line 54:
```suggestion
// Copyright \(C\) 20\d{2} Percona LLC
```

BupycHuk (approved):
LGTM

denisok (approved):

PR #459, opened 2022-03-25
Update mongodb_exporter.service

eval-exec:
Fix type `-` in `EnvironmentFile` section


ShashankSinha252:
Hi @eval-exec , thanks for the PR!

We evaluated the change and came to the conclusion that we would like to keep the service configuration file as is. When MongoDB exporter is installed as a service via SystemD, it is expected that the user will create a configuration file for the exporter. Since the package doesn't install one by default, we decided to not error out in case of missing configuration file.

PR #465, opened 2022-03-31
Pmm 9757 panic on lost connection

ShashankSinha252:
Changed the way we instantiate the topology info object.
Now we call the newTopologyInfo and it never returns an error and the object is never nil. If it is not possible to connect, topology labels remain empty but we return a valid and not nill object.
Also added context termination check to avoid the ugly stack traces produced by context timeouts.

[PMM-9757](https://jira.percona.com/browse/PMM-9757) MongoDB exporter panics when DB connection is lost

### DIFF
--- a/exporter/topology_info.go
+++ b/exporter/topology_info.go
@@ -24,6 +24,7 @@ import (
 	"github.com/percona/percona-toolkit/src/go/mongolib/proto"
 	"github.com/percona/percona-toolkit/src/go/mongolib/util"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"go.mongodb.org/mongo-driver/bson/primitive"
 	"go.mongodb.org/mongo-driver/mongo"
 )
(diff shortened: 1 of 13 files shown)


exporter/topology_info.go
```diff
@@ -62,18 +63,19 @@ type topologyInfo struct {
 // ErrCannotGetTopologyLabels Cannot read topology labels.
 var ErrCannotGetTopologyLabels = fmt.Errorf("cannot get topology labels")
 
-func newTopologyInfo(ctx context.Context, client *mongo.Client) (*topologyInfo, error) {
+func newTopologyInfo(ctx context.Context, client *mongo.Client) *topologyInfo {
 	ti := &topologyInfo{
 		client: client,
 		labels: make(map[string]string),
+		rw:     sync.RWMutex{},
```
JiriCtvrtka, line 70: @ShashankSinha252 This mutex looks ununsed. Or do I missed something?

JiriCtvrtka (approved):

PR #510, opened 2022-06-30
PMM-9919 Add scrape metrics per collector

floatinginbits:
[PMM-9919](https://jira.percona.com/browse/PMM-9919) Add scrape time metrics for all collectors


BupycHuk:
go.mod is not updated, so it's failing

Issue #545, opened 2022-08-17
Failing CI Pipeline: Duplicate Metric Failure


a-gavin:
**Describe the bug**
GitHub action pipelines for various PRs going back several months are failing with errors like output listed below in **Logs** when the changes they test should presumably not cause these issues, e.g. [this one](https://github.com/percona/mongodb_exporter/actions/runs/2870120705).

I have managed to trace the error down to a bug in [`makeMetrics()`](https://github.com/percona/mongodb_exporter/blob/5c3358c3bab3823588fb801a84f3eb7985d2765c/exporter/metrics.go#L241), which is called by the `ServerStatsCollectorTest` via the `ServerStatsCollector` [`collect()`](https://github.com/percona/mongodb_exporter/blob/5c3358c3bab3823588fb801a84f3eb7985d2765c/exporter/serverstatus_collector.go#L54) method.

Effectively what happens is when the collector attempts to take the MongoDB server status output and recursively unpack its Bson format in `makeMetrics()`, the multi-planner histogram data is unpacked as separate metrics with each histogram metric value having the same name causing the error. Documentation on these multi-planner histogram metrics is sparse given how recent MongoDB has implemented them, but information on it can be found in their issue tracker [here](https://jira.mongodb.org/browse/SERVER-63642) and on the commit which added this feature [here](https://github.com/mongodb/mongo/commit/43434627e89822b7e19e3a9d3aeb341be331aae6).

To fix this, these metrics should be logically grouped together and output with additional labels indicating the "bucket" of the histogram the metric corresponds to following the [Prometheus Exposition Format](https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md). Once I figure out how to properly "deserialize" the Bson data passed to `makeMetrics()`, I'll submit a patch which fixes this issue.

**To Reproduce**
Rerun any existing pipelines which fail with the above error output. The ones I've seen fail are:
- 1.17.x, mongo 4.4, ubuntu latest
- 1.17.x, mongo 5.0 ubuntu latest
- 1.17.x, percona/percona-server-mongodb 4.4, ubuntu latest

**Expected behavior**
Histogram metrics should be properly generated according to the [Prometheus Exposition Format](https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md) and TestServerStatusDataCollector should pass.

**Logs**
```console
=== RUN   TestServerStatusDataCollector
    serverstatus_collector_test.go:59: 
[...]

PR #546, opened 2022-08-18
Fix erroring CI duplicate metrics #545

a-gavin:
Separate handling of multi-planner histogram data to account for duplicate values with different "buckets" of data.
Move common makeMetrics code to helper function.

Open to suggestions about how to handle Bson data/Go best practices as I'm unfamiliar. Additionally, in testing there is no option to enable the ServerStatusCollector. I had to manually print out output of `makeRawMetric()` in `makeSingleMetric()` to verify histogram metrics.

### DIFF
--- a/exporter/metrics.go
+++ b/exporter/metrics.go
@@ -18,6 +18,7 @@ package exporter
 
 import (
 	"regexp"
+	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


a-gavin:
Fixed by #547

PR #706, opened 2023-09-06
Export currentOP uptime query metrics

tregubov-av:
This is the implementation of the collector for notifications about too long requests are in processed.
For example:
```
apiVersion: v1
data:
  kube-state-metrics-mongodb.rules: |-
    groups:
    - name: kube-state-metrics-mongodb.rules
      rules:
      - alert: MongodbCurrentQueryTime
        expr: (mongodb_currentop_query_uptime > 3e+8) / 1000
        labels:
          severity: critical
        annotations:
          description: "Opid: {{ $labels.opid }}\nDesc: {{ $labels.desc }}\nNs: {{ $labels.ns }}\nOp : {{ $labels.op }}\nUptime : {{ $value }} ms\n"
          summary: "MongoDB\nCurrent slow query on: {{ $labels.endpoint }}"
kind: ConfigMap
metadata:
  labels:
    app: prometheus
    prometheus: kube-prometheus
    release: kube-prometheus-stack
    role: alert-rules
  name: kube-prometheus-exporter-mongodb
  namespace: monitoring
  ```

---

- [x] Tests passed.
- [x] Fix conflicts with target branch.

### DIFF
--- a/exporter/currentop_collector.go
+++ b/exporter/currentop_collector.go
@@ -1,0 +1,144 @@
+// mongodb_exporter
+// Copyright (C) 2017 Percona LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package exporter
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/sirupsen/logrus"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+type currentopCollector struct {
+	ctx            context.Context
+	base           *baseCollector
+	compatibleMode bool
+	topologyInfo   labelsGetter
+}
+
+var ErrInvalidOrMissingInprogEntry = fmt.Errorf("invalid or misssing inprog entry in currentop results")
+
+// newCurrentopCollector creates a collector for being processed queries.
+func newCurrentopCollector(ctx context.Context, client *mongo.Client, logger *logrus.Logger,
+	compatible bool, topology labelsGetter,
+) *currentopCollector {
+	return &currentopCollector{
+		ctx:            ctx,
+		base:           newBaseCollector(client, logger),
+		compatibleMode: compatible,
+		topologyInfo:   topology,
+	}
+}
+
+func (d *currentopCollector) Describe(ch chan<- *prometheus.Desc) {
+	d.base.Describe(d.ctx, ch, d.collect)
+}
+
+func (d *currentopCollector) Collect(ch chan<- prometheus.Metric) {
+	d.base.Collect(ch)
+}
+
+func (d *currentopCollector) collect(ch chan<- prometheus.Metric) {
+	defer measureCollectTime(ch, "mongodb", "currentop")()
+
+	logger := d.base.logger
+	client := d.base.client
+
+	// Get all requests that are being processed except system requests (admin and local).
+	cmd := bson.D{
+		{Key: "currentOp", Value: true},
+		{Key: "active", Value: true},
+		{Key: "microsecs_running", Value: bson.D{
+			{Key: "$exists", Value: true},
+		}},
+		{Key: "op", Value: bson.D{{Key: "$ne", Value: ""}}},
+		{Key: "ns", Value: bson.D{
+			{Key: "$ne", Value: ""},
+			{Key: "$not", Value: bson.D{{Key: "$regex", Value: "^admin.*|^local.*"}}},
+		}},
+	}
+	res := client.Database("admin").RunCommand(d.ctx, cmd)
+
+	var r primitive.M
+	if err := res.Decode(&r); err != nil {
+		ch <- prometheus.NewInvalidMetric(prometheus.NewInvalidDesc(err), err)
+		return
+	}
+
+	logger.Debug("currentop response from MongoDB:")
+	debugResult(logger, r)
+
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by mihanik712 — APPROVED** [review]

File: exporter/currentop_collector.go [hunk 1]
```diff
@@ -0,0 +1,144 @@
...
+		decs, ok := bsonMapElement["desc"].(string)
+		if !ok {
+			logger.Errorf("Invalid type string assertion for 'desc': %t", ok)
+			break
+		}
+		microsecs_running, ok := bsonMapElement["microsecs_running"].(int64)
+		if !ok {
+			logger.Errorf("Invalid type int64 assertion for 'microsecs_running': %t", ok)
+			break
+		}
```
BupycHuk, line 128: do we really need all these assertions?

BupycHuk (approved):
Thanks @tregubov-av,

Looks good to me.
Could you please sign the CLA?

tregubov-av:
> Thanks @tregubov-av,
> 
> Looks good to me. Could you please sign the CLA?
Please tell me how to do it?
I clicked the subscribe button but the status didn't change

exporter/currentop_collector.go
tregubov-av: I think for a more flexible construction of alerts, you need to have a sufficient number of labels.
In this case, you have to use type assertion to get the values from the interface.
But I don't want to panic if the code fails.
If there are suggestions on how to do better, I'm ready to rewrite this part

tregubov-av:
This is weird...
Locally all tests pass on all available versions MongoDB (4.2, 4.4, 5.0)
What am I doing wrong?
Prepared a pull request based on this documentation https://github.com/percona/mongodb_exporter/blob/main/docs/development-guide.md

BupycHuk:
Hi @tregubov-av, yeah, problem isn't in your PR. We started having this problem for all PRs recently and thanks for inputs. we will try to fix that.

exporter/currentop_collector.go
```diff
@@ -0,0 +1,144 @@
...
+)
+
+type currentopCollector struct {
+	ctx            context.Context
+	base           *baseCollector
+	compatibleMode bool
+	topologyInfo   labelsGetter
+}
+
+var ErrInvalidOrMissingInprogEntry = fmt.Errorf("invalid or misssing inprog entry in currentop results")
```
artemgavrilov, line 37:
```suggestion
var ErrInvalidOrMissingInprogEntry = errors.New("invalid or missing inprog entry in currentop results")
```

exporter/currentop_collector.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	for _, bsonMap := range inprog {
+
+		bsonMapElement, ok := bsonMap.(primitive.M)
+		if !ok {
+			logger.Errorf("Invalid type primitive.M assertion for bsonMap: %t", ok)
+			break
+		}
+		opid, ok := bsonMapElement["opid"].(int32)
```
artemgavrilov, line 103: Are we sure that field `opid` always exists? If not, then we can get panic here on casting `nil` to `int32`.

Same comment valid for all other fields that we checking in this file.

tregubov-av: I think if the process starts executing, then it will definitely have an identifier.
I implemented filtering by fields that may be missing at the request level

```
	cmd := bson.D{
		{Key: "currentOp", Value: true},
		{Key: "active", Value: true},
		{Key: "microsecs_running", Value: bson.D{
			{Key: "$exists", Value: true},
		}},
		{Key: "op", Value: bson.D{{Key: "$ne", Value: ""}}},
		{Key: "ns", Value: bson.D{
			{Key: "$ne", Value: ""},
			{Key: "$not", Value: bson.D{{Key: "$regex", Value: "^admin.*|^local.*"}}},
		}},
	}
```

artemgavrilov: Ok, let's keep it as is.

exporter/currentop_collector.go
```diff
@@ -0,0 +1,144 @@
...
+	if !ok {
+		ch <- prometheus.NewInvalidMetric(prometheus.NewInvalidDesc(ErrInvalidOrMissingInprogEntry),
+			ErrInvalidOrMissingInprogEntry)
+	}
+
+	for _, bsonMap := range inprog {
+
+		bsonMapElement, ok := bsonMap.(primitive.M)
+		if !ok {
+			logger.Errorf("Invalid type primitive.M assertion for bsonMap: %t", ok)
```
artemgavrilov, line 100: I guess it should be like this:
```suggestion
			logger.Errorf("Invalid type primitive.M assertion for bsonMap: %t", bsonMapElement)
```

Same for other log messages.

artemgavrilov: I think `%T` will fit better here, because we interested what was original type and why it was unexpected for us.

exporter/currentop_collector.go
```diff
@@ -0,0 +1,144 @@
...
+		ch <- prometheus.NewInvalidMetric(prometheus.NewInvalidDesc(ErrInvalidOrMissingInprogEntry),
+			ErrInvalidOrMissingInprogEntry)
+	}
+
+	for _, bsonMap := range inprog {
+
+		bsonMapElement, ok := bsonMap.(primitive.M)
+		if !ok {
+			logger.Errorf("Invalid type primitive.M assertion for bsonMap: %T", bsonMapElement)
+			break
```
artemgavrilov, line 101: And the last comment from me, it looks like it makes sense to use here `continue` instead of `break`.

artemgavrilov: Here and in other places.

tregubov-av: As far as I understand, all elements in the map will have the same structure, and if the types do not match in at least one, then we can be sure that the types will not match in the remaining elements of the map.
Does it make sense to continue execution thereby creating a burden due to unsuccessful operations?

artemgavrilov: Hmm...🤔
Well `happy path` should not reach this error at all, but if one object was corrupted for some reason it still makes sense to check other and return at least something.

tregubov-av: Okay, I agree that it's better to get something than nothing.
I'll make changes now PR

tregubov-av: I`am done :)

artemgavrilov: Great!)

artemgavrilov (approved):

BupycHuk (approved):

tregubov-av:
👍

PR #752, opened 2023-12-01
Support dynamic target

taoso:
In my product environment, there are too many targets to feed them into the `mongodb.uri` option. Please support create target dynamically according to the scraping parameter.

### DIFF
--- a/exporter/multi_target_test.go
+++ b/exporter/multi_target_test.go
@@ -67,6 +67,39 @@ func TestMultiTarget(t *testing.T) {
 
 	// Test all targets
 	for sn, opt := range opts {
-		assert.HTTPBodyContains(t, multiTargetHandler(serverMap), "GET", fmt.Sprintf("?target=%s", opt.URI), nil, expected[sn])
+		assert.HTTPBodyContains(t, multiTargetHandler(serverMap, nil), "GET", fmt.Sprintf("?target=%s", opt.URI), nil, expected[sn])
+	}
+}
+
+func TestDynamicTarget(t *testing.T) {
+	hostname := "127.0.0.1"
+
+	hosts := []string{
+		fmt.Sprintf("mongodb://%s", net.JoinHostPort(hostname, tu.GetenvDefault("TEST_MONGODB_STANDALONE_PORT", "27017"))),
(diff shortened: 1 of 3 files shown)


taoso:
It seems the failing of CI jobs caused by unsuccessful booting of docker compose. Can anyone make some help?

I can run the `make test` on my mac machine.

image
image

BupycHuk:
Hi, sorry, I don't think that it's the way we want to go.

taoso:
> Hi, sorry, I don't think that it's the way we want to go.

Hi, @BupycHuk , Do you kinda to offer any help? The ci failed before go test.

image

taoso:
Hi,

The failed checks still are relevant  to `docker compose up -d --wait`. However, I make no change to this part. Besides, the check for 5.0 and percona 5.0 works. I do not know what should I to to fix this issue. Can anybody make any help?

taoso:
Hello @JiriCtvrtka @BupycHuk Could you kind to offer any advice to help me fix the CI problem? The error log shows that both test for mongo:4.4 and percona:4.4 are failed at `container mongo-rs1-setup exited (1)`. It does not run the go test at all. What can I do to fix this issue?

Thanks.

JiriCtvrtka:
@taoso Hi. Right now I am working on fixing CI. Thanks for your contribution. I will message you once problem will be fixed.

BupycHuk:
Hi, this approach might let anyone use someones mongodb_exporter to DDoS other MongoDB Server, so it's better to have preconfigured list of Servers to monitor them in multi-target option.

taoso:
> Hi, this approach might let anyone use someones mongodb_exporter to DDoS other MongoDB Server, so it's better to have preconfigured list of Servers to monitor them in multi-target option.

Hi @BupycHuk please reconsider this PR.

Although PR makes mongodb_expoter support multiple target dynamically, it does not means anyone can DDoS other MongoDB dynamically because there is still need to set the username and password on the exporter side.

I am working for monitoring huge number of mongodb instances, besides all of them will be registered into CMDB dynamically. It's not easy to preconfigured all list of servers in the exporter and update them dynamically.

Please reconsider this PR. Thank you.

4orty:
@JiriCtvrtka @BupycHuk 
Is there no plan to support this feature? Actually, many other exporters support this feature and it seems like this feature is a requirement to achieve complete multi-target support. cc. @taoso

PR #780, opened 2024-01-17
Fix splitting cluster URI into multiple targets

adnull:
As kong option parser doesn't have an option not to split quoted arguments by comma and just strips all quotes from arguments, there is no chance to specify comma-containing cluster URIs as they will be treated as separated multiple targets ( https://github.com/percona/mongodb_exporter/pull/653#issuecomment-1894399492 ). 

This version improves URI splitting algorythm as folows:
- URI prefixed my mongodb:// or mongodb+srv:// can contain comma-separated hosts without schema part and will be treated as a cluster. Any other separated host or a cluster must by prefixed by mongodb:// to be treated as a separate target.
- If no one host in the line is prefixed by scheme part, all they will be treated as separate targets

- [ ] Links to related pull requests (optional).

---


main.go
```diff
@@ -168,20 +169,43 @@ func buildExporter(opts GlobalFlags, uri string, log *logrus.Logger) *exporter.E
...
+func parseURIList(uriList []string) []string {
+	var URIs []string
+
+	// If server URI is prefixed with mongodb, then every next URI in line not prefixed with mongodb is a part of cluster
+	// Otherwise treat it as a standalone server
+	realURI := ""
+	matchRegexp := regexp.MustCompile(`^mongodb(\+srv)?://`)
+
+	for idx := range uriList {
+		URI := uriList[idx]
```
BupycHuk, line 190:
```suggestion
	for idx, URI := range uriList {
```

main.go
```diff
@@ -168,20 +169,43 @@ func buildExporter(opts GlobalFlags, uri string, log *logrus.Logger) *exporter.E
...
+				realURI = realURI + "," + URI
+			}
+		} else {
+			if realURI != "" {
+				URIs = append(URIs, realURI)
+			}
+			realURI = URI
+		}
+		if idx == len(uriList)-1 && realURI != "" {
+			URIs = append(URIs, realURI)
```
BupycHuk, line 205: I think we can just move it after the loop without if condition. as far as I see `realURI` can't be empty.

main.go
```diff
@@ -168,20 +169,43 @@ func buildExporter(opts GlobalFlags, uri string, log *logrus.Logger) *exporter.E
...
 		}
-
-		servers[serverIdx] = buildExporter(opts, URI, log)
 	}
-
-	return servers
+	return URIs
```
BupycHuk, line 208: please fix this

main_test.go
```diff
@@ -16,12 +16,27 @@
...
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 )
 
+func TestParseURIList(t *testing.T) {
```
BupycHuk, line 26: I think we can add t.Parallel() here

main_test.go
```diff
@@ -16,12 +16,27 @@
...
+	tests := map[string][]string{
+		"server1":                 {"mongodb://server1"},
+		"server1,server2,server3": {"mongodb://server1", "mongodb://server2", "mongodb://server3"},
+		"mongodb.server,server2":  {"mongodb://mongodb.server", "mongodb://server2"},
+		"standalone,mongodb://server1,server2,mongodb+srv://server3,server4,mongodb://server5": {"mongodb://standalone", "mongodb://server1,server2", "mongodb+srv://server3,server4", "mongodb://server5"},
+	}
```
BupycHuk, line 32: let's add a few more cases with `mongodb://` in the beginning of the string

BupycHuk (approved):
LGTM

main_test.go
adnull: Indeed, done

JiriCtvrtka (approved):

Issue #763, opened 2023-12-15
Using environment variables, startup errors 


xushiba:
**Describe the bug**
When using multi-target collection, starting with an environment variable will report an error

**To Reproduce**
export MONGODB_USER=XXX
export MONGODB_PASSWORD=YYY
mongodb_exporter --mongodb.uri=mongodb://127.0.0.1:17001,mongodb://127.0.0.1:17001 

**Expected behavior**
But using this approach it's possible to：
mongodb_exporter --mongodb.uri=mongodb://user:pwd@127.0.0.1:17001,mongodb://user:pwd@127.0.0.1:17001 

**Logs**
Cannot connect to MongoDB: invalid dsn: error parsing uri: scheme must be \"mongodb\" or \"mongodb+srv\""

pflong:
@xushiba What's your mongodb exporter version ?

xushiba:
> @xushiba What's your mongodb exporter version ?

0.40

xushiba:
Follow-up: MONGODB_USER varibales cannot be set to mongodb-xxx

PR #800, opened 2024-02-20
Fix case where users are prefixed with mongodb

bvalente:
In issue https://github.com/percona/mongodb_exporter/issues/763 a user faced an error when connecting to the uri.

There is a bug when the mongodb user is prefixed with 'mongodb' that prevents the program to add the `mongodb://` uri prefix.

### DIFF
--- a/main.go
+++ b/main.go
@@ -207,7 +207,6 @@ func parseURIList(uriList []string) []string {
 }
 
 func buildURI(uri string, user string, password string) string {
-
 	prefix := "mongodb://" // default prefix
 	matchRegexp := regexp.MustCompile(`^mongodb(\+srv)?://`)
 
--- a/main.go
+++ b/main.go
@@ -207,17 +207,26 @@ func parseURIList(uriList []string) []string {
 }
 
 func buildURI(uri string, user string, password string) string {
+
+	prefix := "mongodb://" // default prefix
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -215,7 +215,7 @@ func buildURI(uri string, user string, password string) string {
 		// add user and pass to the uri
 		uri = fmt.Sprintf("%s:%s@%s", user, password, uri)
 	}
-	if !strings.HasPrefix(uri, "mongodb") {
+	if !strings.HasPrefix(uri, "mongodb://") {
```
BupycHuk, line 218:
```suggestion
	if !strings.HasPrefix(uri, "mongodb://") &&  !strings.HasPrefix(uri, "mongodb+srv://") {
```

BupycHuk: could you add to test cases origin which contains mongodb+srv prefix?

BupycHuk:
@bvalente thank you for your contribution, seems like your changes won't work for cases with mongodb+srv prefixes

main.go
bvalente: Hello @BupycHuk, I pushed a new commit that addresses considerations for URIs with the srv prefix, thank you for the review

bvalente:
Hello again @BupycHuk, the last build failed because of a lint check, I pushed a new commit to amend the lint

BupycHuk (approved):

JiriCtvrtka (approved):

PR #818, opened 2024-03-15
PMM-12161 fix rs state wrong metric

igroene:
[PMM-12161](https://jira.percona.com/browse/PMM-12161) 

- Fixes exporter returning status 6 when it cannot read the replica set state. This was a bad design decision and causes mongos routers to return the mongodb_mongod_replset_my_state even though that metric does not make sense for a router. 

[PMM-12161]: https://perconadev.atlassian.net/browse/PMM-12161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BupycHuk (approved):

BupycHuk (changes requested):
Hi @igroene this change leads to panic in our tests and I'm sure that the same behavior will be on production

Issue #773, opened 2024-01-04
SRV with direct connect


ariksidney:
**Is your feature request related to a problem? Please describe.**
We're using Atlas on AWS with a private endpoint. When I try to run the exporter with the `srv` connection string, I cannot connect to the MongoDB cluster. However if use a non `srv` connection string (with all the replica set nodes in it) and use `--mongodb.direct-connect`, the connection works without any issues.

**Describe the solution you'd like**
A solution would be to resolve the `srv` dns entry if the `--mongodb.direct-connect` option is set and the uri is a `srv` connection string. A new exporter would then be created for each replica.

**Describe alternatives you've considered**
The easiest alternative would be to just use the replica nodes uri instead of the `srv` connection string. However we currently only have the `srv` connection string in our configMaps.

**Additional context**
If you think this feature would be helpful for the project, I would gladly contribute it.

PR #835, opened 2024-04-03
Serving SRV targets and getting all nodes in single request

adnull:
As it stated in https://github.com/percona/mongodb_exporter/issues/773 there is no chance to monitor cluster nodes with the only mongodb+srv url specified. 
According to the specification, direct connection couldn't be used with a cluster, specified as a SRV domain

This update adds 
- --split-cluster option, which:
  - extracts mongodb+srv domain into cluster connection string containing nodes from DNS server
  - splits all cluster connection string into separate targets that can be queried by "target" parameter of multitarget endpoint
- overall target endpoint "/scrapeall" which outputs all the targets metrics with a "instance" label containing nodename as a host:port pair (or only host if no port specified)

Combining all that features we can specify cluster url as mongodb+srv domain and monitor all the nodes metrics and states in one request by querying /scrapeall endpoint


JiriCtvrtka (approved):

exporter/server.go
```diff
@@ -97,6 +104,68 @@ func multiTargetHandler(serverMap ServerMap) http.HandlerFunc {
 	}
 }
 
+// OverallTargetsHandler is a handler to scrape all the targets in one request.
+// Adds instance label to each metric.
+func OverallTargetsHandler(exporters []*Exporter, logger *logrus.Logger) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		seconds, err := strconv.Atoi(r.Header.Get("X-Prometheus-Scrape-Timeout-Seconds"))
+		// To support also older ones vmagents.
```
idoqo, line 112:
```suggestion
		// To support older vmagents.
```

exporter/server.go
```diff
@@ -97,6 +104,68 @@ func multiTargetHandler(serverMap ServerMap) http.HandlerFunc {
...
+		if err != nil {
+			seconds = 10
```
idoqo, line 114: Helpful to log this? So folks don't scratch their heads when their set timeout isn't working.

exporter/server.go
```diff
@@ -97,6 +104,68 @@ func multiTargetHandler(serverMap ServerMap) http.HandlerFunc {
...
+		}
+
+		var gatherers prometheus.Gatherers
+		gatherers = append(gatherers, prometheus.DefaultGatherer)
+
+		filters := r.URL.Query()["collect[]"]
+
+		for _, e := range exporters {
+			ctx, cancel := context.WithTimeout(r.Context(), time.Duration(seconds-e.opts.TimeoutOffset)*time.Second)
+			defer cancel()
```
idoqo, line 124: These defer (and the one on L135) won't run when each loop ends though, is that intentional?

adnull: As I understand the logic we create gatherers in the loop, but real metrics collecting is being done by promhttp.HandlerFor which calls collector.Collect and it is located outside the loop, so we free the resources after the all work done.

exporter/server.go
adnull: logged that as a debug message

idoqo (approved):

REFERENCE.md
```diff
@@ -1,34 +1,34 @@
...
+|--collector.currentopmetrics|Enable collecting metrics from currentop admin command|
+|--collector.indexstats|Enable collecting metrics from $indexStats|
+|--collector.collstats|Enable collecting metrics from $collStats|
+|--collect-all|Enable all collectors. Same as specifying all --collector.\<name\>|
+|--collector.collstats-limit=0|Disable collstats, dbstats, topmetrics and indexstats collector if there are more than \<n\> collections. 0=No limit|
+|--collector.profile-time-ts=30|Set time for scrape slow queries| This interval must be synchronized with the Prometheus scrape interval|
+|--collector.profile|Enable collecting metrics from profile|
+|--collector.shards|Enable collecting metrics related to Mongo shards|
+|--metrics.overridedescendingindex| Enable descending index name override to replace -1 with _DESC ||
+|--version|Show version and exit|
```
BupycHuk, line 34: `--collector.pbm` is lost

BupycHuk (approved):

REFERENCE.md
adnull: damn, I checked the list after merge but missed it. fixed.

exporter/exporter.go
```diff
@@ -386,6 +354,42 @@ func (e *Exporter) Handler() http.Handler {
...
+		case "indexstats":
+			requestOpts.EnableIndexStats = true
+		case "collstats":
+			requestOpts.EnableCollStats = true
+		case "profile":
+			requestOpts.EnableProfile = true
+		case "shards":
+			requestOpts.EnableShards = true
+		case "pbm":
+			requestOpts.EnablePBMMetrics = true
```
idoqo, line 386: seems FCV was missed during the merge:
```suggestion
		case "pbm":
			requestOpts.EnablePBMMetrics = true
		case "fcv":
			requestOpts.EnableFCV = true
		
```

idoqo (changes requested):
Looks good, we should also enable fcv collector.

exporter/exporter.go
adnull: Omg, how did that happen.. fixed

idoqo (approved):

PR #863, opened 2024-06-17
PMM-13141 Added feature compatibility version collector

cfsalguero:
[PMM-13141](https://jira.percona.com/browse/PMM-13141 

Added a new collector for Feature compatibility version,

[PMM-13141]: https://perconadev.atlassian.net/browse/PMM-13141?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


exporter/exporter.go
```diff
@@ -70,6 +70,9 @@ type Opts struct {
 	EnableCollStats          bool
 	EnableProfile            bool
 	EnableShards             bool
+	EnableFCV                bool // Feature Compatibility Version.
+
+	FCVInterval time.Duration
```
BupycHuk, line 75: I'm not sure if we need separate interval for FCV.

exporter/feature_compatibility_version_collector.go
```diff
@@ -0,0 +1,103 @@
...
+		if err := res.Decode(&m); err != nil {
+			ch <- prometheus.NewInvalidMetric(prometheus.NewInvalidDesc(err), err)
+			return
+		}
+
+		d.lastScrape = d.now()
+
+		rawValue := walkTo(m, []string{"featureCompatibilityVersion", "version"})
+		if rawValue != nil {
+			version, err := strconv.ParseFloat(fmt.Sprintf("%v", rawValue), 64)
```
BupycHuk, line 90: wouldn't it better to use hashicorps version library to parse it?

BupycHuk (commented):
LGTM, just few comments

exporter/exporter.go
igroene: this value changes only after upgrades so it should have a big interval to avoid unnecessary reads

BupycHuk: that's true, but it's not a good practice in exporter development. Is this operation generate a big load on system?

igroene: no, not really. We can use one of the default collection intervals

BupycHuk (approved):

PR #356, opened 2021-10-05
Expose exporter.Handler and allow disabling default registry

rfratto:
The Grafana Agent (https://github.com/grafana/agent) embeds mongodb_exporter as an integration, where it needs to be able to call into the exporter's HTTP handler for its all-in-one server. This endpoint also needs to only return metrics for the exporter itself and not from the global registry, which is populated by unrelated metrics.

Signed-off-by: Robert Fratto <robertfratto@gmail.com>

PR #885, opened 2024-08-05
Add command line flag to disable metrics about the exporter it-self

equinox0815:
PR #356 introduced an API option to disable default exporter metrics (process_* and go_*). However there is no command line flag to actually set this option. This change adds the command line flag `--no-collector.exporter-metrics` to disable those metrics.

### DIFF
--- a/main.go
+++ b/main.go
@@ -49,6 +49,7 @@ type GlobalFlags struct {
 	LogLevel              string   `name:"log.level" help:"Only log messages with the given severity or above. Valid levels: [debug, info, warn, error, fatal]" enum:"debug,info,warn,error,fatal" default:"error"`
 	ConnectTimeoutMS      int      `name:"mongodb.connect-timeout-ms" help:"Connection timeout in milliseconds" default:"5000"`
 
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -49,6 +49,7 @@ type GlobalFlags struct {
 	LogLevel              string   `name:"log.level" help:"Only log messages with the given severity or above. Valid levels: [debug, info, warn, error, fatal]" enum:"debug,info,warn,error,fatal" default:"error"`
 	ConnectTimeoutMS      int      `name:"mongodb.connect-timeout-ms" help:"Connection timeout in milliseconds" default:"5000"`
 
+	DisableExporterMetrics   bool `name:"web.disable-exporter-metrics" help:"Exclude metrics about the exporter itself (process_*, go_*)"`
```
BupycHuk, line 52: I believe it should be `--no-collector.exporter-metrics`

equinox0815: Yeah this is probably more in line with the naming the exporter uses. I was not sure which name is preferred so i was taking inspiration from the node_exporter. I will change the name and update the PR.

BupycHuk (approved):

JiriCtvrtka (approved):

PR #891, opened 2024-08-16
Fix/index stats exclude system collections

ramunas-omnisend:
Details:
When a view is created inside a MongoDB database, a special collection named `system.views` is also created. Attempting to call `IndexStats` on `system.views` results in an `Unauthorized` error, even when executed with root privileges.

For example:

```bash
ERRO[0011] cannot get $indexStats cursor for collection <database>.system.views: 
(Unauthorized) not authorized on <database> to execute command { 
aggregate: "system.views", pipeline: [ { $indexStats: {} } ], cursor: {}, 
lsid: { id: UUID("<lsid>") }, $clusterTime: { clusterTime: Timestamp(<timestamp>, 1), 
signature: { hash: BinData(0, <signature>), keyId: <key-id> } }, $db: "<database>", 
$readPreference: { mode: "primaryPreferred" } }
```

This fix filters out system collections like `system.views` when gathering `IndexStats`, preventing these unnecessary errors from appearing in the logs.


ramunas-omnisend:
Will open another PR

PR #892, opened 2024-08-16
Fix/ignore views for list collections

ramunas-omnisend:
When listCollections is executed, it may encounter views, which can cause errors such as:

```bash
ERRO[0016] cannot get $collstats cursor for collection <database>.<view>: (CommandNotSupportedOnView) Namespace <database>.<collection> is a view, not a collection
```

This fix filters out views when listing collections, preventing these errors from appearing in the logs.


ramunas-omnisend:
will open another PR

PR #1080, opened 2025-05-09
Cache prometheusize rename between scrapes

vigno88:
After profiling the exporter because it had a somewhat high cpu usage, it reported that the regexp operations in this function accounts for more than 10% of the cpu time. 

More precisely, the first call to prometheusize uses 3.79%, the second 5.86% (in exporter.MakeMetrics) and lastly it 1.03% (in exporter.MakeRawMetrics)

This all goes to zero if we cache them.

### DIFF
--- a/exporter/metrics.go
+++ b/exporter/metrics.go
@@ -18,6 +18,7 @@ package exporter
 import (
 	"regexp"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
@@ -164,14 +165,14 @@ var (
 	dollarRe              = regexp.MustCompile(`\_$`)
 )
 
-var prometheusizeCache = make(map[string]string)
+var prometheusizeCache = sync.Map{}
 
 // prometheusize renames metrics by replacing some prefixes with shorter names
 // replace special chars to follow Prometheus metric naming rules and adds the
 // exporter name prefix.
 func prometheusize(s string) string {
-	if renamed, exists := prometheusizeCache[s]; exists {
-		return renamed
+	if renamed, exists := prometheusizeCache.Load(s); exists {
+		return renamed.(string)
 	}
 	backup := strings.Clone(s)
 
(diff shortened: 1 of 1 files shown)


vigno88:
[pprof.mongodb_exporter.samples.cpu.pb.gz](https://github.com/user-attachments/files/20127121/pprof.mongodb_exporter.samples.cpu.pb.gz)

exporter/metrics.go
```diff
@@ -164,10 +164,17 @@ var (
...
 )
 
+var prometheusizeCache = make(map[string]string)
+
 // prometheusize renames metrics by replacing some prefixes with shorter names
 // replace special chars to follow Prometheus metric naming rules and adds the
 // exporter name prefix.
 func prometheusize(s string) string {
+	if renamed, exists := prometheusizeCache[s]; exists {
+		return renamed
```
idoqo, line 174: This is missing the exporter prefix.

idoqo (changes requested):
Thanks a lot for your contribution. Please
- Fix failing tests.
- Sign the CLA

idoqo:
Hi @vigno88 looks like tests are still failing - also looks like the profile you shared was from before the change - will be nice to have both to see how they compare.

vigno88:
Hi, thanks for the quick response.

I'll take a look at the tests and yes i will provide a new profile.

Nathan AV

BupycHuk:
Hi, please also make sure that it's thread-safe.

vigno88:
There are other changes is this profile, but this show the cpu usage of prometheusize after the changes.

vigno88:
All the test have passed, and the signature for my account has passed (vigno99 is the user of my computer)

idoqo:
Thanks a lot for your contribution, @vigno88. I'll check with our internal QA team about testing this, and then we can merge.

idoqo (approved):

idoqo:
Hi @vigno88 , ready to merge this now, but seems there were also some commits from @vigno99 , so the CLA needs to be signed by both accounts

PR #1083, opened 2025-05-12
PMM-13959 remove GSSAPI from main builds

idoqo:
[PMM-13748](https://jira.percona.com/browse/PMM-13748)

[PMM-13959](https://jira.percona.com/browse/PMM-13959)

[PMM-13748]: https://perconadev.atlassian.net/browse/PMM-13748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PMM-13959]: https://perconadev.atlassian.net/browse/PMM-13959?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -3,6 +3,25 @@ version: 2
 builds:
   - binary: mongodb_exporter
     env:
+    goos:
+      - linux
+      - darwin
+      - windows
+    goarch:
+      - amd64
+      - arm
+      - arm64
+    goarm:
+      - 7
+    ignore:
+      - goos: darwin
+        goarch: arm
+    flags:
+      - -v
+    ldflags:
+        - -s -w -X main.version=v{{.Version}} -X main.commit={{.ShortCommit}} -X main.buildDate={{.Date}}
+  - binary: mongodb_exporter.gssapi
+    env:
       - CGO_ENABLED=1
     goos:
       - linux
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ init:                       ## Install linters
 	cd tools && go generate -x -tags=tools
 
 build:                      ## Compile using plain go build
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -71,7 +71,7 @@ init:                       ## Install linters
 	cd tools && go generate -x -tags=tools
 
 build:                      ## Compile using plain go build
-	CGO_ENABLED=1 go build -ldflags="$(GO_BUILD_LDFLAGS)"  -o $(PMM_RELEASE_PATH)/mongodb_exporter -tags gssapi
+	go build -ldflags="$(GO_BUILD_LDFLAGS)"  -o $(PMM_RELEASE_PATH)/mongodb_exporter
```
BupycHuk, line 74: this revert PR could include only this as everything else doesn't affect PMM really

idoqo: We want to still keep GSSAPI support in goreleaser builds? @BupycHuk

idoqo: I imagine we can drop it too, and in that case drop the tests as well (which removes the need for the other files).

We still have the original PR, so if we need to prepare some builds, we can base them on that.

BupycHuk: We can have additional build of mongodb_exporter with gssapi

ademidoff (approved):

.goreleaser.yml
```diff
@@ -2,6 +2,25 @@ dist: build
 version: 2
 builds:
   - binary: mongodb_exporter
+    env:
+    goos:
+      - linux
+      - darwin
+      - windows
```
ademidoff, line 9: I wouldn't care about windows :)

PR #1094, opened 2025-06-08
PMM-13881: Use VERSION file for exporter version tag

idoqo:
[PMM-13881](https://jira.percona.com/browse/PMM-13881)

[PMM-13881]: https://perconadev.atlassian.net/browse/PMM-13881?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BupycHuk:
Have you checked if goreleaser uses the same file?

idoqo:
> Have you checked if goreleaser uses the same file?

It uses the tags instead (https://goreleaser.com/cookbooks/using-main.version/) - so I think we will still do releases, PMM builds (or builds generated with `make build`) will just rel on the VERSION file instead

.github/workflows/version-check.yml
```diff
@@ -0,0 +1,75 @@
...
+          
+          if ! semver "$GIT_VERSION_CLEAN" >/dev/null 2>&1; then
+            echo "ERROR: Git tag contains invalid semantic version: $GIT_VERSION"
+            exit 1
+          fi
+          
+          # Compare versions: -1 if first < second, 0 if equal, 1 if first > second
+          COMPARISON=$(semver -r ">=$GIT_VERSION_CLEAN" "$FILE_VERSION_CLEAN" && echo "valid" || echo "invalid")
+          
+          if [ "$COMPARISON" = "invalid" ]; then
```
ademidoff, line 69:
```suggestion    
          if [ ! $(semver -r ">=$GIT_VERSION_CLEAN" "$FILE_VERSION_CLEAN") ]; then
```

ademidoff: nit: this would be simpler, but still do the same

ademidoff (approved):

PR #1142, opened 2025-08-12
PMM-7: Fix disabled CGO_ENABLED in release builds

idoqo:


BupycHuk (approved):
Can you add CI job which runs on main?

.github/workflows/go.yml
```diff
@@ -66,6 +66,10 @@ jobs:
           env_vars: GO_VERSION,TEST_MONGODB_IMAGE
           fail_ci_if_error: false
 
+      - name: Dry-run GoReleaser
+        run: |
+          make release-dry-run
+
```
BupycHuk, line 72: it will run on every matrix element which isn't great, let's make another job and rename current one to test instead of build

idoqo: yeah reworked them now.

Issue #1153, opened 2025-08-21
Installation from .deb fails on Ubuntu


zbjornson:
**Describe the bug**

**To Reproduce**
1. On Ubuntu (others?) `sudo apt install ./mongodb_exporter-0.47.0.linux-64-bit.deb`. Eventually it fails with this:
```
Setting up mongodb_exporter (0.47.0) ...
Creating user and group...
Unknown option: comment
adduser [--home DIR] [--shell SHELL] [--no-create-home] [--uid ID]
...
```

**Expected behavior**
Install completes, adding a user.

**Environment**
 - Ubuntu 20.04.6 LTS

**Additional context**
Bug was introduced by d128386a5f5df0918ab46a5bf88f8060fe82564a in v0.45.0.

PR #1182, opened 2025-10-16
fix: use useradd instead of adduser for cross-version compatibility

izmal:
Replace adduser with useradd for mongodb_exporter user creation. The --comment option in adduser is not supported in Debian 11, only in Debian 12+. useradd works consistently across all versions.

related https://github.com/percona/mongodb_exporter/issues/1153


ademidoff:
Hi @izmal!

Could you please sign the CLA (see link above)?

PR #1195, opened 2025-11-14
PMM-14431 More complex check if it is real count type or not.

JiriCtvrtka:
PMM-14431

FB: https://github.com/Percona-Lab/pmm-submodules/pull/4118


JiriCtvrtka:
> I'm seeing `GetAllIndexesForCollections` being called in multiple collectors. This call is rather expensive. Would it be beneficial to cache it for a short period of time?

Rewritten as we discussed on Go BE.

ademidoff (approved):

PR #1210, opened 2025-12-03
Fix /dev/null redirection typo in systemctl command

guoard:


guoard:
Closed in favour of #1212

PR #1280, opened 2026-05-06
PMM-15047 Update base image to alpine 3.23.4 with specific SHA

radoslawszulgo:
It's recommended to pin the version of base images inside of Docker containers. Using a dynamic version can cause unexpected behavior and at worst, can lead to supply chain attacks. On top of that the 'latest' tag does not always automatically refer to the newest version of the image, so it can also lead to using an outdated version of the base image.

re: https://perconadev.atlassian.net/browse/PMM-15047


maxkondr (approved):

ademidoff:
We are not shipping the alpine image - it’s only used to build the exporter, while scratch is used to ship it.

Therefore, there is no point in pinning the version, it will only create a maintenance tax.

PR #1295, opened 2026-05-28
Fix inverted mongodb_mongos_sharding_chunks_is_balancer_running gauge

nisarg14:
**Fix inverted mongodb_mongos_sharding_chunks_is_balancer_running gauge**

**Summary**

- Fixes chunksBalancerRunning() so mongodb_mongos_sharding_chunks_is_balancer_running is 1 when MongoDB balancerStatus.inBalancerRound is true, and 0 otherwise.

- Previously the condition was negated (!m.InBalancerRound), so the gauge reported the opposite of the metric name and help text (“Shard balancer is in a balancing round”).

- Renames the existing TestMongosMetrics subtest to match what it actually tests (balancer_enabled via mode).

- Adds a subtest that exercises chunksBalancerRunning() against inBalancerRound from balancerStatus.


nisarg14:
Please add `pmm-review-exporters` as reviewers per CONTRIBUTING.md

exporter/v1_compatibility_test.go
```diff
@@ -245,6 +245,39 @@ func TestMongosMetrics(t *testing.T) {
...
+		var metric prometheus.Metric
+		var m dto.Metric
+		metric, err = chunksBalancerRunning(ctx, client)
+		assert.NoError(t, err)
+
+		err = metric.Write(&m)
+		assert.NoError(t, err)
+
+		expected := 0
+		if bs.InBalancerRound {
```
ademidoff, line 309: This `expected` computation mirrors the function under test exactly (`chunksBalancerRunning` applies the same `if inBalancerRound { 1 }` mapping), so the assertion is tautological. It does guard against a re-introduced `!` inversion, but on an idle test cluster `inBalancerRound` is effectively always false, so only the `→ 0` path is ever exercised — the `→ 1` path (the behavior this PR fixes) is never covered. Consider extracting the bool→value mapping into a pure function and unit-testing both inputs, or at minimum add a comment noting this limitation.

ademidoff (commented):
Confirmed the production fix is correct. The inversion is a regression from PMM-12962 (`15688e3`), which renamed `chunksBalanced`/`is_balanced` ("Shards are balanced", correctly `!inBalancerRound`) to `chunksBalancerRunning`/`is_balancer_running` but kept the `!m.InBalancerRound` logic, inverting the new semantics. This restores the intended behavior (the metric is only emitted under `--compatible-mode`). One note on the new test below.

ademidoff:
> Please add `pmm-review-exporters` as reviewers per CONTRIBUTING.md

That group was removed, so we are good - `CONTRIBUTING.md` is updated and we have reviewers assigned :)

exporter/v1_compatibility_test.go
nisarg14: Thanks for the feedback. I’ve now extracted the bool->value mapping into a pure function and added a unit test covering both true and false cases. The integration test is kept only to verify the MongoDB response->metric wiring.

exporter/v1_compatibility_test.go
```diff
@@ -211,9 +211,39 @@ func TestCreateOldMetricFromNew(t *testing.T) {
...
+		},
+		{
+			name:            "balancer not running",
+			inBalancerRound: false,
+			expected:        0,
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
```
ademidoff:
```suggestion
```
It's redundant in modern Go.

exporter/v1_compatibility_test.go
ademidoff: Awesome. One final fix - let's run `make format`, commit and push - it should make the linter happy :)

nisarg14: Sure, i've ran `make format` and pushed the code

nisarg14: Gentle reminder @ademidoff i've ran the command can you please review

nisarg14: @ademidoff @JiriCtvrtka gentle reminder !

JiriCtvrtka (approved):

nisarg14:
@JiriCtvrtka can you help here, i see checks are failing

JiriCtvrtka:
@nisarg14 Done. Let’s wait for Alex, since he requested changes earlier. I think we can merge this PR without a QA check.

nisarg14:
@ademidoff can you help here

nisarg14:
@ademidoff gentle reminder for review

PR #1321, opened 2026-07-21
PMM-15198 Indexstats fix.

JiriCtvrtka:
PMM-15198

FB: https://github.com/Percona-Lab/pmm-submodules/pull/4507

**Summary**
mongodb_indexstats_* metrics are silently dropped when the exporter scrapes a sharded cluster through mongos.

**Steps to reproduce**
Run a sharded MongoDB cluster (2+ shards) with a sharded collection that has at least one index.
Point mongodb_exporter at mongos with --collector.indexstats enabled.
Scrape /metrics.

**Actual result**
Only one shard's index statistics are exposed. The exporter logs duplicate-metric errors from the Prometheus registry, and the remaining shards' data is discarded.

**Expected result**
One series per index per shard, each distinguished by a shard label.

**Root cause**
When scraping via mongos, the $indexStats aggregation returns one document per shard for every index. The indexstats collector built its label set per document but never read the shard field, so every shard produced an identical metric descriptor. Prometheus rejects duplicate descriptors, so all but one shard was thrown away.

A related defect exists in the collstats collector: the labels map is allocated once before the document loop, so a shard value from one document leaks into subsequent documents that report no shard.

**Fix**
Add a shared setShardLabel() helper in exporter/common.go that sets the shard label unconditionally (empty string when absent), so all series of a metric family carry identical label dimensions and no registration conflict occurs.
Call it from both the indexstats and collstats collectors.
Move the collstats label map allocation inside the document loop to stop cross-document leakage.
Impact on existing series

Non-sharded deployments see no change — Prometheus drops empty label values at ingestion. Sharded deployments gain one series per shard per index instead of losing them.

**Testing**
Unit tests for setShardLabel() (shard present / empty / missing / non-string). Integration tests TestIndexStatsCollectorSharded and TestCollStatsCollectorSharded assert metrics from ≥2 shards with no duplicates. Note: these integration tests currently skip because the compose test cluster registers no shards — a separate test-infrastructure issue, tracked independently of this fix.


JiriCtvrtka:
@copilot review

JiriCtvrtka:
@copilot review

exporter/indexstats_collector.go
```diff
@@ -144,6 +141,17 @@ func (d *indexstatsCollector) collect(ch chan<- prometheus.Metric) {
 	}
 }
 
+func indexStatsLabels(labels map[string]string, database, collection, indexName string, metric bson.M) map[string]string {
+	labels["database"] = database
+	labels["collection"] = collection
+	labels["key_name"] = indexName
+	if shard, ok := metric["shard"].(string); ok && shard != "" {
```
ademidoff: Minor: the empty-string guard makes indexstats behave differently from collstats, which adds the label on `ok` alone (`collstats_collector.go:151`). Either the guard matters — then collstats should get it too — or it can't happen and it's dead code here. Since it's a two-line change under the same ticket, consider applying the same guard to collstats in this PR.

Context: label dimensions must be consistent per metric name within a single scrape, so if a mongos ever returned `shard: ""` for some documents but real names for others, dropping the label on only those series would make the whole `mongodb_indexstats_*` family fail to gather. Theoretical, but worth keeping the two collectors aligned.

exporter/indexstats_collector_test.go
```diff
@@ -85,6 +85,59 @@ mongodb_indexstats_accesses_ops{collection="testcol_02",database="testdb",key_na
...
+		"shard":      "shard-0",
+	}, first)
+	assert.Equal(t, map[string]string{
+		"cl_role":    "mongos",
+		"database":   "testdb",
+		"collection": "orders",
+		"key_name":   "_id_",
+		"shard":      "shard-1",
+	}, second)
+	assert.NotEqual(t, first, second)
```
ademidoff: Minor: this assertion is redundant — the two exact-map assertions above already guarantee `first != second`.

exporter/indexstats_collector_test.go
```diff
@@ -85,6 +85,59 @@ mongodb_indexstats_accesses_ops{collection="testcol_02",database="testdb",key_na
...
+
+func TestIndexStatsLabelsWithoutShard(t *testing.T) {
+	t.Parallel()
+
+	labels := indexStatsLabels(
+		map[string]string{},
+		"testdb",
+		"orders",
+		"_id_",
+		bson.M{"shard": ""},
```
ademidoff: This only covers `shard: ""`. The more common real case is the field being entirely absent (any non-mongos node) — consider adding a `bson.M{}` input as well.

ademidoff (commented):
A few minor suggestions, nothing blocking.

exporter/indexstats_collector.go
JiriCtvrtka: Good catch. Applied your primary suggestion: kept the && shard != "" guard in indexstats and added the same guard to collstats (collstats_collector.go:151), so both collectors now behave identically.

exporter/indexstats_collector_test.go
JiriCtvrtka: Removed.

exporter/indexstats_collector_test.go
JiriCtvrtka: Added test case.

JiriCtvrtka:
@copilot review

exporter/collstats_collector.go
```diff
@@ -159,4 +157,11 @@ func (d *collstatsCollector) collect(ch chan<- prometheus.Metric) {
 	}
 }
 
+func setShardLabel(labels map[string]string, metrics bson.M) {
```
JiriCtvrtka: Although PMM-15198 primarily targets indexstats, collstats uses the same optional shard label and was updated to keep both collectors consistent. Since collstats reuses the labels map across results, an empty or absent shard could otherwise retain the previous shard value and mislabel metrics. The added tests cover both cases.

JiriCtvrtka:
@copilot review

JiriCtvrtka:
@copilot review

4nte (approved):

exporter/common.go
```diff
@@ -269,6 +269,25 @@ func splitNamespace(ns string) (string, string) {
...
+// without a shard would silently inherit it.
+//
+// Empty and absent shards are both treated as "no shard". Whether the label is
+// present has to be the same for every document of a metric name within one
+// scrape: the descriptors are built from these labels, and a registry rejects a
+// collector describing one fully-qualified name with two different label sets,
+// which makes the whole scrape fail. That holds in practice because $collStats
+// and $indexStats report a shard for every document through mongos and for none
+// of them otherwise.
+func setShardLabel(labels map[string]string, doc bson.M) {
```
ademidoff, line 278: Two notes on the invariant this comment documents.

**The failure mechanism is a panic at registration, not a `Gather` error.** Worth being precise, because it changes how you'd debug a report. `makeRegistry` calls `registry.MustRegister(ic)` (`exporter/exporter.go:207`) on every scrape from inside the HTTP handler (`exporter/exporter.go:363`), and `Describe` runs the full collect. `Registry.Register` returns `"descriptors reported by collector have inconsistent label names or help strings for the same fully-qualified name"` when one fqName is described with two label sets, so `MustRegister` panics, `net/http` recovers, and the scrape returns 500. Note the handler's `ErrorHandling: promhttp.ContinueOnError` does not help here: it only covers `Gather`, which happens after registration.

**Residual risk: mixed shard presence within one scrape.** The claim that mongos reports a shard for every document holds for user collections, but explicit-namespace mode bypasses the `systemDBs` filter (`exporter/common.go:32`). So `--mongodb.collstats-colls=config.something` on mongos routes to the config server rather than a shard; if that path returns no `shard` field while user collections do, the entire scrape 500s -- a hard-to-diagnose failure triggered by an unrelated flag value.

Cheap hardening that removes the invariant entirely:

```go
func setShardLabel(labels map[string]string, doc bson.M) {
	shard, _ := doc["shard"].(string)
	labels["shard"] = shard
}
```

Label dimensions become unconditionally consistent, and Prometheus treats an empty label value as equivalent to an absent label at ingestion, so non-sharded deployments see no series change. That also lets you drop the `delete` and most of this comment. This risk is pre-existing for collstats and not a regression from this PR, so folding it in is your call -- but this is the natural moment.

Separately, 12 lines of doc for a 4-line function is heavy. The invariant paragraph is genuinely worth keeping (it's non-obvious and load-bearing); the rest could be two lines.

exporter/indexstats_collector.go
```diff
@@ -144,6 +141,16 @@ func (d *indexstatsCollector) collect(ch chan<- prometheus.Metric) {
 	}
 }
 
+// indexStatsLabels builds the label set for a single $indexStats document.
+func indexStatsLabels(labels map[string]string, database, collection, indexName string, metric bson.M) map[string]string {
```
ademidoff: This wrapper doesn't quite earn its keep. It takes a map, mutates it, and returns the same map, while the doc comment says it "builds the label set" -- and it's inconsistent with `collstats`, which calls `setShardLabel` directly (`exporter/collstats_collector.go:151`). As far as I can tell it exists only to make a trivial thing testable.

Dropping it makes the collector change a single added line and matches the sibling collector:

```go
labels := d.topologyInfo.baseLabels()
labels["database"] = database
labels["collection"] = collection
labels["key_name"] = indexName
setShardLabel(labels, metric)
```

That also removes the need for two of the three new test functions.

exporter/indexstats_collector_test.go
```diff
@@ -82,7 +82,81 @@ mongodb_indexstats_accesses_ops{collection="testcol_02",database="testdb",key_na
 		"mongodb_indexstats_accesses_ops",
 	}
 	err := testutil.CollectAndCompare(c, expected, filter...)
-	assert.NoError(t, err)
+	require.NoError(t, err)
+}
+
+func TestIndexStatsLabels(t *testing.T) {
```
ademidoff: Three test functions now cover one 4-line helper, with overlapping cases: `TestSetShardLabel` (`exporter/common_test.go:192`) covers set/empty/absent, and `TestIndexStatsLabelsWithoutShard` (`exporter/indexstats_collector_test.go:122`) covers absent/empty again one layer up.

`TestIndexStatsLabels` specifically asserts that two calls with two separate input maps produce two separate results, which is trivially true for any function that mutates its argument -- it can't fail, and it isn't the bug being fixed. I'd keep `TestSetShardLabel` and delete both of these along with the `indexStatsLabels` wrapper.

exporter/indexstats_collector_test.go
```diff
@@ -82,7 +82,81 @@ mongodb_indexstats_accesses_ops{collection="testcol_02",database="testdb",key_na
 		"mongodb_indexstats_accesses_ops",
 	}
 	err := testutil.CollectAndCompare(c, expected, filter...)
-	assert.NoError(t, err)
+	require.NoError(t, err)
```
ademidoff: The actual regression -- one collection, multiple shard documents, distinct series emitted -- has no test. The infrastructure exists: `tu.DefaultTestClientMongoS` (`internal/tu/testutils.go:76`), and `exporter/exporter_test.go:65` already scrapes through mongos on port 17000. A test that builds an `indexstatsCollector` against the mongos client and asserts the exposed series carry `shard="..."` (ideally after `shardCollection`, so two documents come back) would cover the fix and also catch the registration panic described in my comment on `exporter/common.go`. That's the test I'd trade all three label tests for.

Unrelated to this: the `assert` -> `require` conversions inside the insert/index loops in this file are a genuine improvement (fail fast instead of spamming), but the trailing `assert.Equal` -> `require.Equal` swaps in `TestSanitize` change nothing and just add noise to the diff.

ademidoff (commented):
Two things that don't map to a line:

**Stale PR description.** The second paragraph ("raises the `go test` timeout from 30s to 1m in the Makefile") no longer applies: `main` already carries `-timeout 1m` (`Makefile:124`, `:127`, `:130`) and no Makefile change appears in this diff. Commit `bf207f6b` is in the branch history but contributes nothing against the current base. Worth trimming so reviewers aren't hunting for it.

**Cardinality, for the release notes.** The duplicate indexstats series were previously deduplicated silently by `Gather`; now they're all exposed. Scraping mongos in discovering mode goes from `collections x indexes` to `collections x indexes x shards` series. That's the correct data, but it's a real increase users should be told about.

exporter/indexstats_collector.go
JiriCtvrtka: Gone. indexstats now calls `setShardLabel(labels, metric)` directly after `labels["key_name"]`, exactly as you wrote it, so the collector change is one added line and matches the sibling collector.

exporter/common.go
JiriCtvrtka: All three taken.

**Mechanism.** Corrected — the comment now names `MustRegister` panicking when one fqName is described with two label sets, instead of implying a `Gather` error. Worth recording that both modes apply here and they are distinct: inconsistent label *dimensions* fail registration, duplicate *identical* series fail in `Gather`. The reported bug is the second, your residual risk is the first.

**Hardening.** Adopted your snippet as-is. The `delete` is gone and the label is set unconditionally, so label dimensions are consistent by construction — which also closes the explicit-namespace path you describe, where `--mongodb.collstats-colls=config.something` on mongos could return no shard while user collections do.

**Doc length.** Down from 12 lines to 6, keeping only the paragraph on why the label is unconditional.

exporter/indexstats_collector_test.go
JiriCtvrtka: Done — `TestIndexStatsLabels` and `TestIndexStatsLabelsWithoutShard` are deleted along with the wrapper. `TestSetShardLabel` stays and covers four inputs: shard set, empty string, field absent, field not a string. It pre-seeds the map with `shard: "shard-0"` before each case, so it asserts the overwrite rather than the assignment — which is the collstats leak, tested directly.

exporter/indexstats_collector_test.go
JiriCtvrtka: Both addressed, and one disclosure.

**The test.** Added `TestIndexStatsCollectorSharded` and `TestCollStatsCollectorSharded`, built the way you described: `tu.DefaultTestClientMongoS`, `shardCollection` with a hashed `_id` so the initial chunks presplit across shards and each returns a document, collector registered on a `NewPedanticRegistry`, asserting no index exposes the same shard twice. `require.NoError` on `Gather` is the direct assertion of the reported symptom — before the fix it fails with `was collected before with the same name and label values` — and `Register` covers the registration panic from your `common.go` note. Common setup is extracted into `gatherShardedMetrics` and `shardedSeriesLabels`.

**`assert` -> `require`: you were right and I resolved it wrong the first time.** I reverted the swaps in `TestSanitize` but kept the trailing ones after `CollectAndCompare`, which is the line this comment is anchored on and the same "changes nothing" case, since `FailNow` has nothing left to skip at the end of a function. Those are reverted now, plus one in a deferred `Drop`. What remains is only the conversions inside the insert and index-creation loops.

**The disclosure: these two tests do not currently execute, anywhere.** The compose cluster registers no shards — `listShards` returns `[]` and `sh.addShard` fails for both sets with `FailedToSatisfyReadPreference`. CI proved it is not just my machine: I briefly pushed a variant that fails instead of skipping, and `c430822` with `t.Skipf` was 11/11 green while `118859d` with `t.Fatalf` came back 3 green and 8 red across the image matrix. So these tests have been skipping in CI since the moment they were added.

Two reasons it went unnoticed. `setup-shard.sh` pipes its commands into the client as a heredoc, so they run as JavaScript and the client exits 0 even when all of them failed — a cluster with no shards is indistinguishable from a successful setup, and `restart: on-failure` never triggers. And CI has only a fixed `sleep 10` after `docker compose up -d`, with no check that any shard registered.

I am back on `t.Skipf` so this PR is not gated on infrastructure, and the tests stay so the coverage lands the moment the fixture works. But that means what you asked for is satisfied on paper and not in effect, and I would rather say so than leave a green PR with a test that never runs.

**Your call on the fix, since it is a topology decision rather than a bug.** rs1 runs as `--shardsvr` with no auth; rs2 runs `mongod --replSet rs2 --port 27017 --oplogSize 16 --auth --keyFile=/opt/keyfile` — no `--shardsvr`, auth on, while mongos has neither auth nor a keyfile. If that reading is right the fixture can provide at most one shard, and multi-shard coverage needs two by definition. So either the compose topology changes, or these two tests get a dedicated sharded cluster via testcontainers the way `percona_info_schema_process_list_integration_test.go` does in mysqld_exporter — though that file's `//go:build integration` tag means it never runs in CI, so it would have to run untagged to be worth anything. Separate ticket either way, and I have the diagnosis written up.

JiriCtvrtka:
@copilot review

ademidoff (approved):

PR #1347, opened 2026-08-25
Redact MongoDB URI credentials before logging them

### LINKED ISSUE #1345: MongoDB connection URIs with embedded credentials are logged in cleartext
**Describe the bug**

Several log statements write MongoDB connection URIs verbatim. Because the exporter injects the monitoring user's credentials into the URI before these statements run, the emitted log line contains `user:password` in cleartext.

This came up while reviewing #1336, which redacts one of these call sites as a side effect of its feature work. The remaining sites are unrelated to that PR and should be fixed separately — hence this issue.

**Affected call sites** (all on `main` @ 9585218)

| # | Location | Level | Reachable at default log level |
|---|---|---|---|
| 1 | `main.go:143` — `log.Debug("Connection URI", "uri", uri)` | Debug | No (needs `--log.level=debug`) |
| 2 | `exporter/seedlist.go:31` — `log.Fatalf("Failed to parse URI %s: %v", uri, err)` | Fatal | Yes |
| 3 | `exporter/seedlist.go:36` — `logger.Error("Failed to lookup SRV records", "uri", uri, ...)` | Error | Yes |
| 4 | `exporter/seedlist.go:41` — `logger.Error("No SRV records found", "uri", uri)` | Error | Yes |
| 5 | `main.go:259` — `log.Fatalf("Failed to parse URI %s: %v", hosturl, err)` (the `--split-cluster` path) | Fatal | Yes |
| 6 | `exporter/server.go:181` — `log.Error("Unable to parse provided address as url", "address", e.opts.URI, ...)` | Error | Yes |

**Why the URI carries credentials**

`buildURI` (`main.go:300`) injects them explicitly:

```go
if parsedURI.User == nil && user != "" && password != "" {
    parsedURI.User = url.UserPassword(user, password)
}
```

and the fallback path `buildURIManually` (`main.go:291`) does the same by string concatenation. A URI supplied directly via `--mongodb.uri` / `MONGODB_URI` may of course already embed credentials, which covers the seedlist and `buildServerMap` cases where `buildURI` isn't involved.

**Severity notes**

- Site 1 only fires with debug logging enabled. That is not the default, but debug logs are exactly what users attach to support tickets and GitHub issues, so credentials leak into third-party hands routinely.
- Sites 2–6 fire at the **default** log level, on error paths that are easy to hit in practice: a malformed URI, an unresolvable SRV record, or a DNS hiccup. Sites 3 and 4 in particular are plain operational failures, not misconfiguration — an SRV lookup failure on a `mongodb+srv://` URI will print the credentials on every affected startup.
- These are the more serious of the two groups, and none of them are covered by #1336.

**Expected behavior**

No log statement should emit a MongoDB URI with a readable password, at any level.

**Suggested fix**

`(*url.URL).Redacted()` handles the cases where the URI parses:

```go
func redactMongoURI(rawURI string) string {
    uri, err := url.Parse(rawURI)
    if err != nil {
        return "<invalid MongoDB URI>"
    }
    return uri.Redacted()
}
```

Two things to watch when applying it:

- Sites 2, 5, and 6 log *because* parsing failed, so `Redacted()` is not available on the parsed value — those need string-level scrubbing (or should log only the host portion / omit the URI entirely and rely on the error).
- A shared helper is preferable to per-site fixes so new logging doesn't reintroduce the leak. #1336 adds an unexported `redactMongoURI` in `package main`; `exporter/seedlist.go` and `exporter/server.go` live in `package exporter` and cannot reuse it as written. Placing it somewhere both packages can import (e.g. `internal/`) would be worthwhile.

**Related**

- #1336 — redacts site 1 only, as incidental cleanup. Sites 2–6 are out of scope there.

**Environment**

Not environment-specific; applies to all builds from `main`.

ademidoff:
Fixes #1345.

MongoDB connection URIs carrying credentials were written to the logs verbatim. `buildURI` (`main.go:300`) injects the monitoring user's password into the URI, and a URI supplied via `--mongodb.uri`/`MONGODB_URI` may already embed one, so several log statements disclosed `user:password` in cleartext.

## What changed

A new `internal/redact` package holds the two helpers. The leaking call sites straddle `package main` and `package exporter`, so neither could host a shared helper; `internal/util` is MongoDB-command code and pulls in the driver, which made it a poor home for a pure string function.

- `redact.MongoURI(uri)` — `(*url.URL).Redacted()` on the happy path, keeping the username and substituting `xxxxx`.
- `redact.Error(err)` — scrubs credentials out of an error's message (see below).

Six call sites now redact:

| Location | Level |
|---|---|
| `main.go:143` — `"Connection URI"` | Debug |
| `main.go:259` — `--split-cluster` parse failure | Fatal |
| `exporter/seedlist.go:31` — URI parse failure | Fatal |
| `exporter/seedlist.go:36` — SRV lookup failure | Error |
| `exporter/seedlist.go:41` — no SRV records | Error |
| `exporter/server.go:181` — `buildServerMap` parse failure | Error |

Only the first is Debug-level; the other five fire at the **default** log level, on error paths reachable through ordinary operational failures such as a DNS hiccup or a malformed URI.

## The error argument leaked too

Three of those sites log *because* `url.Parse` failed. `url.Parse` returns a `*url.Error`, whose `Error()` quotes the offending URL **verbatim** — so the `err` argument alone disclosed the full password, and redacting only the URI argument would have fixed nothing there. Hence `redact.Error`.

This is not hypothetical. The URI PMM generates with an escaped socket path is exactly what makes `url.Parse` fail:

```
before: address=mongodb://monitor:SuperSecret123@%2F... error="parse \"mongodb://monitor:SuperSecret123@%2F...\": invalid URL escape \"%2F\""
after:  address=mongodb://monitor:xxxxx@%2F...         error="parse \"mongodb://monitor:xxxxx@%2F...\": invalid URL escape \"%2F\""
```

Where parsing failed there is no `*url.URL` to redact, so `MongoURI` falls back to a regexp that rewrites the userinfo of the raw string. The URI is kept in the message rather than dropped, because with multiple targets configured it is the only way to tell which one is malformed.

## Tests

`internal/redact/redact_test.go` is table-driven over credentialed, uncredentialed, percent-encoded, SRV, multi-host and unparseable inputs. The assertion that matters is that the password substring does not survive. `TestErrorHidesURIQuotedByParseError` first asserts that the raw `url.Parse` error really does contain the password, then that the redacted form does not — so the test would catch a future Go release that changed this behaviour.

Also verified manually against the built binary for three scenarios (credentials via `--mongodb.user`/`--mongodb.password`, credentials embedded in `--mongodb.uri`, and the socket URI), grepping the logs for the password in each.

## Not changed

`buildURI` and `buildURIManually` are untouched — injecting the credentials is intended; only logging them was wrong.

Two spots were examined and left alone: `exporter/pbm_collector.go:88` passes the URI to the third-party PBM SDK and logs only `err`, and `exporter/exporter.go:426` wraps a driver validation error whose messages quote individual option keys rather than the whole URI. Both carry residual third-party risk outside this repo's control.

Pre-existing lint findings in the touched files are left in place to keep the diff surgical. `golangci-lint run --new-from-rev=origin/main` reports 0 issues, which matches what CI's reviewdog gate (`filter_mode: added`) evaluates.

## Conflict with #1336

PR #1336 adds an unexported `redactMongoURI` to `package main` and redacts `main.go:143` as incidental cleanup. Whichever lands second needs a small rebase there — drop the local helper in favour of `internal/redact`, or re-apply that one line. No other overlap.

### DIFF
--- a/exporter/pbm_collector.go
+++ b/exporter/pbm_collector.go
@@ -27,6 +27,7 @@ import (
 	"go.mongodb.org/mongo-driver/mongo"
 
 	"github.com/percona/mongodb_exporter/internal/proto"
+	"github.com/percona/mongodb_exporter/internal/redact"
 	"github.com/percona/mongodb_exporter/internal/util"
 )
 
@@ -87,7 +88,7 @@ func (p *pbmCollector) collect(ch chan<- prometheus.Metric) {
 	pbmEnabledMetric := 0
 	pbmClient, err := sdk.NewClient(p.ctx, p.mongoURI)
 	if err != nil {
-		logger.Warn("failed to create PBM client", "error", err.Error())
+		logger.Warn("failed to create PBM client", "error", redact.Error(err))
 		return
 	}
 	defer func() {
@@ -136,7 +137,8 @@ func (p *pbmCollector) collect(ch chan<- prometheus.Metric) {
 func (p *pbmCollector) pbmAgentMetrics(ctx context.Context, pbmClient *sdk.Client, l *slog.Logger, currentNode *proto.HelloResponse) []prometheus.Metric {
 	clusterStatus, err := cli.ClusterStatus(ctx, pbmClient, cli.RSConfGetter(p.mongoURI))
 	if err != nil {
-		l.Error("failed to get cluster status", "error", err.Error())
+		// PBM wraps the URI it was handed into this error, so it cannot be logged as it comes.
+		l.Error("failed to get cluster status", "error", redact.Error(err))
 		return nil
 	}
 
--- a/exporter/seedlist.go
+++ b/exporter/seedlist.go
@@ -30,7 +30,9 @@ import (
 func GetSeedListFromSRV(uri string, logger *slog.Logger) string { //nolint:cyclop
 	uriParsed, err := url.Parse(uri)
 	if err != nil {
-		log.Fatalf("Failed to parse URI %s: %v", redact.MongoURI(uri), redact.Error(err))
+		// The parse error is not logged: it quotes the URI cut short at the byte url.Parse choked
+		// on, and when that byte is "?" or "#" from the password, the quote keeps the part before.
+		log.Fatalf("Failed to parse URI %s", redact.MongoURI(uri))
 	}
 
 	cname, srvRecords, err := net.LookupSRV("mongodb", "tcp", uriParsed.Hostname())
--- a/exporter/seedlist.go
+++ b/exporter/seedlist.go
@@ -22,23 +22,25 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+
+	"github.com/percona/mongodb_exporter/internal/redact"
 )
 
 // GetSeedListFromSRV converts mongodb+srv URI to flat connection string.
 func GetSeedListFromSRV(uri string, logger *slog.Logger) string { //nolint:cyclop
 	uriParsed, err := url.Parse(uri)
 	if err != nil {
-		log.Fatalf("Failed to parse URI %s: %v", uri, err)
+		log.Fatalf("Failed to parse URI %s: %v", redact.MongoURI(uri), redact.Error(err))
 	}
 
 	cname, srvRecords, err := net.LookupSRV("mongodb", "tcp", uriParsed.Hostname())
 	if err != nil {
-		logger.Error("Failed to lookup SRV records", "uri", uri, "error", err)
+		logger.Error("Failed to lookup SRV records", "uri", redact.MongoURI(uri), "error", err)
 		return uri
 	}
 
 	if len(srvRecords) == 0 {
-		logger.Error("No SRV records found", "uri", uri)
+		logger.Error("No SRV records found", "uri", redact.MongoURI(uri))
 		return uri
 	}
 
--- a/exporter/server.go
+++ b/exporter/server.go
@@ -28,6 +28,8 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/exporter-toolkit/web"
+
+	"github.com/percona/mongodb_exporter/internal/redact"
 )
 
 // ServerMap stores http handlers for each host
@@ -178,7 +180,7 @@ func buildServerMap(exporters []*Exporter, log *slog.Logger) ServerMap {
 		if parsedURL, err := url.Parse(e.opts.URI); err == nil {
 			servers[parsedURL.Host] = e.Handler()
 		} else {
-			log.Error("Unable to parse provided address as url", "address", e.opts.URI, "error", err)
+			log.Error("Unable to parse provided address as url", "address", redact.MongoURI(e.opts.URI), "error", redact.Error(err))
 		}
 	}
 
--- a/exporter/server.go
+++ b/exporter/server.go
@@ -180,7 +180,10 @@ func buildServerMap(exporters []*Exporter, log *slog.Logger) ServerMap {
 		if parsedURL, err := url.Parse(e.opts.URI); err == nil {
 			servers[parsedURL.Host] = e.Handler()
 		} else {
-			log.Error("Unable to parse provided address as url", "address", redact.MongoURI(e.opts.URI), "error", redact.Error(err))
+			// The parse error is not logged: it quotes the URI cut short at the byte url.Parse
+			// choked on, and when that byte is "?" or "#" from the password, the quote keeps the
+			// part before.
+			log.Error("Unable to parse provided address as url", "address", redact.MongoURI(e.opts.URI))
 		}
 	}
 
--- a/internal/redact/redact.go
+++ b/internal/redact/redact.go
@@ -1,0 +1,52 @@
+// mongodb_exporter
+// Copyright (C) 2025 Percona LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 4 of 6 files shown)


JiriCtvrtka (approved):
Looks good. 

Two places in exporter/pbm_collector.go I think are worth another look. Both are outside this diff, so I couldn't leave inline comments.

```exporter/pbm_collector.go:90 — logger.Warn("failed to create PBM client", "error", err.Error())```
The Not changed section excludes this one because it "logs only err". But that's the case redact.Error was added for — per The error argument leaked too, redacting only the URI argument "would have fixed nothing there". Same shape here: the URI goes into sdk.NewClient, and pbm/connect.Connect wraps a *url.Error that renders it back out.

```exporter/pbm_collector.go:139 — l.Error("failed to get cluster status", "error", err.Error())```
This one isn't mentioned. cli.ClusterStatus(ctx, pbmClient, cli.RSConfGetter(p.mongoURI)) hands the URI to PBM, and at the pinned version (percona-backup-mongodb@v1.8.1-0.20251124214042-d06cab743541, sdk/cli/status.go:168-180):
```
curi, err := url.Parse(string(g))
if err != nil {
    return nil, errors.Wrapf(err, "parse mongo-uri '%s'", g)
}
```
g is the full URI. It fails on exactly the escaped-socket-path URI this PR uses as its motivating example, so the password appears twice in the line — once inside the *url.Error, once in PBM's own '%s'. Unlike :90, this fires on every scrape at Error level whenever --collector.pbm is on.

On "residual third-party risk outside this repo's control" — that's true of who builds the message, but redact.Error(err) scrubs it at the log site regardless of origin, which is what makes it work for the url.Parse cases already covered.

Both look like one-line changes:
```
logger.Warn("failed to create PBM client", "error", redact.Error(err))
l.Error("failed to get cluster status", "error", redact.Error(err))
```
Could be there's a reason these differ from the six you converted — happy to be corrected.

**Comment by ademidoff** [comment]
Thanks — you were right about `pbm_collector.go`, and the "logs only `err`" line in my *Not changed* section was a bad argument. Whether the message is safe depends on what the callee puts in it, not on which argument we pass. Both sites are converted in b8baabd. The two differ in what I could actually reproduce, so recording that here.

## `pbm_collector.go:139` — confirmed, exactly as you described

Calling `cli.RSConfGetter(uri).Get(...)` at the pinned version with the escaped-socket URI:

```
parse mongo-uri 'mongodb://monitor:SuperSecret123@%2Ftmp%2Fmongodb.sock/admin': parse "mongodb://monitor:SuperSecret123@%2Ftmp%2Fmongodb.sock/admin": invalid URL escape "%2F"
```

The password twice — once in PBM's own `'%s'`, once in the wrapped `*url.Error` — at `Error` level, on every scrape with `--collector.pbm`. That is the worst of the sites this PR has touched, and it was not in the issue.

## `pbm_collector.go:90` — I could not reproduce a leak

Six malformed URIs through `sdk.NewClient`, all with credentials, none disclosed them:

| URI defect | error |
|---|---|
| escaped socket path | `ping: server selection error: ... Addr: /tmp/mongodb.sock ...` |
| `?ssl=notabool` | `invalid connection string option: error parsing uri: invalid value for "ssl": "notabool"` |
| `host:notaport` | `... port must be an integer: strconv.Atoi: parsing "notaport"` |
| no host | `... must have at least 1 host` |
| raw space / `#` in password | `ping: server selection error: ... dial tcp: lookup host: no such host` |
| `mongodb+srv://` | `... unescaped colon in password` |

That matches the source: `Connect` reaches `errors.Wrap(err, "parse mongo-uri")` at `pbm/connect/connect.go:245` with no `%s`, and everything before it comes from the driver's connstring parser, which names the offending option rather than echoing the URI.

Converted it anyway. Not because I found a leak, but because "we handed this function a credential-bearing URI, so we redact what it hands back" is an invariant that survives a dependency bump, whereas "I read PBM 1.8.1's error strings and they looked fine" has to be re-audited every time. It is one line.

## What `redact.Error` does and does not fix there

It removes the full password from both copies:

```
red = parse mongo-uri 'mongodb://monitor:xxxxx@%2Ftmp%2Fmongodb.sock/admin": invalid URL escape "%2F"
```

It does not close the truncation gap in the wrapped `*url.Error`. For a password containing `?` or `#`, `url.Parse` cuts the URI at that byte, so the inner quote reads `parse "mongodb://monitor:pa"` with no `@` left to anchor on and the prefix `pa` survives. Nothing in the pattern can tell `monitor:pa` from `host:27017`.

So the two groups are handled differently, deliberately:

- **Where we build the message** (the three `url.Parse` sites in `main.go`, `exporter/seedlist.go`, `exporter/server.go`) the parse error is now dropped entirely and only the redacted URI is logged. That closes the gap rather than narrowing it, and costs little — the redacted URI already says which target failed. This also resolves the Copilot comment on this PR, which flagged the same thing.
- **Where PBM builds it**, dropping the error would throw away the only diagnostic, so `redact.Error` stays and the prefix residue remains a known limit. `TestErrorLeavesTruncatedURIAlone` pins it, and `TestErrorRedactsURIWrappedByThirdParty` covers PBM's double-quoted shape.

That is also what keeps `redact.Error` in the tree: dropping the error at the three parse sites left it with no callers, and these two sites are now its reason to exist.

`golangci-lint run --new-from-rev=origin/main` is clean; `internal/redact`, `exporter/dsn_fix` and root tests pass. `make check` and `go test ./...` still fail the same way they do on `main` — 347 pre-existing lint findings, and the `./exporter` / `internal/tu` suites needing `make test-cluster`.

Issue #697, opened 2023-08-11
Member replication lag metric can report wrong value when nodes go away


nick-jones:
**Describe the bug**

Specifically the `mongodb_mongod_replset_member_replication_lag` metric, which is only exposed if compatible mode is enabled with v2.

The wrong lag value is basically due to some part of the lag calculation being zero, resulting in the value being the current unix timestamp (the absolute value of the result is used, so either side of the substraction could be zero).

When the lag is being reported incorrectly, the `state` label is set to `(not reachable/healthy)` - so overall this is easy to mitigate. Equally it's possible to just not use this metric considering it's only available via compatible mode - the calculations can be done via `mongodb_rs_members_optimeDate` instead. But either way, it does seem like a potential bug that the lag is falsely reported as high as it is.

**To Reproduce**

*docker-compose.yaml*
```yaml
version: "3.9"
services:
  mongo0:
    image: mongo:6
    container_name: mongo0
    command: --bind_ip_all --replSet=rs
    ports:
      - "27000:27017"
    expose:
      - "27017"
    volumes:
      - /data/db
  mongo0_exporter:
    image: percona/mongodb_exporter:2.37.0
    container_name: mongo0_exporter
    command:
      - --mongodb.uri=mongodb://mongo0:27017
      - --discovering-mode
      - --collect-all
      - --compatible-mode
      - --web.listen-address=:8080
    ports:
      - "18080:8080"
  mongo1:
    image: mongo:6
    container_name: mongo1
    command: --bind_ip_all --replSet=rs
    expose:
      - "27017"
    volumes:
      - /data/db
  mongo1_exporter:
    image: percona/mongodb_exporter:2.37.0
    container_name: mongo1_exporter
    command:
      - --mongodb.uri=mongodb://mongo1:27017
      - --discovering-mode
      - --collect-all
      - --compatible-mode
      - --web.listen-address=:8080
    ports:
      - "18081:8080"
  mongo2:
    image: mongo:6
    container_name: mongo2
    command: --bind_ip_all --replSet=rs
    expose:
      - "27017"
    volumes:
      - /data/db
  mongo2_exporter:
    image: percona/mongodb_exporter:2.37.0
    container_name: mongo2_exporter
    command:
      - --mongodb.uri=mongodb://mongo2:27017
      - --discovering-mode
      - --collect-all
      - --compatible-mode
      - --web.listen-address=:8080
    ports:
      - "18082:8080"
    expose:
      - "18082"
```

[...]

PR #1363, opened 2026-09-07
Fix compatibility replication lag sign for stale primary optimes

### LINKED ISSUE #1362: Compatibility replication lag reports false positives when secondary optime is newer than cached primary
**Describe the bug**

In compatibility mode, `mongodb_mongod_replset_member_replication_lag` reports a positive lag when a secondary's applied optime is newer than the primary optime in the same `replSetGetStatus` response.

The calculation in [`replSetMetrics`](https://github.com/percona/mongodb_exporter/blob/49408b0b904b1b34b4e0924f85bf7ebe6b2e76fb/exporter/v1_compatibility.go#L1040-L1045) uses:

```go
math.Abs(float64(memberOptime.Unix() - primaryOptime.Unix()))
```

MongoDB documents that information about other replica-set members comes from heartbeats and can be several seconds old: [replSetGetStatus](https://www.mongodb.com/docs/manual/reference/command/replSetGetStatus/). When scraping a secondary, its own optime can therefore be newer than its cached view of the primary. Taking the absolute value turns that negative lag into a false positive, which can trigger replication-lag alerts.

**To Reproduce**

1. Enable compatibility metrics with `--compatible-mode` and replica-set status collection with `--collector.replicasetstatus`.
2. Scrape a secondary whose local applied optime is newer than the heartbeat-reported primary optime.
3. Observe that the compatibility lag metric reports the timestamp difference as positive.

The calculation can also be reproduced deterministically, without a MongoDB connection, by passing this synthetic BSON fixture to `replSetMetrics`:

```go
bson.M{
    "set": "rs0",
    "members": bson.A{
        bson.M{
            "name": "primary:27017", "stateStr": "PRIMARY", "health": 1,
            "optimeDate": primitive.DateTime(1_000_000),
        },
        bson.M{
            "name": "secondary:27017", "stateStr": "SECONDARY", "health": 1,
            "self": true, "optimeDate": primitive.DateTime(1_010_000),
        },
    },
}
```

The timestamps are synthetic BSON dates in milliseconds: primary at Unix second 1000, secondary at second 1010. Both members are healthy and both timestamps are nonzero.

**Expected behavior**

Report `max(0, primaryOptime - memberOptime)`: this fixture should emit zero lag. A secondary genuinely behind the reported primary optime must still emit the positive difference. Keep the existing behavior of omitting lag when no primary is found or either optime is unavailable.

This correction does not remove the underlying heartbeat staleness of remote-member data; it prevents reversing the meaning of the timestamp difference.

**Logs**

Deterministic regression result against unmodified `main`:

```text
mongodb_mongod_replset_member_replication_lag{name="secondary:27017",self="1",set="rs0",state="SECONDARY"}
actual:   10
expected: 0
```

The same test preserves positive lag of 17 and 60 seconds, zero lag when caught up, and omission when the primary or either optime is unavailable.

**Environment**

- Observed with standalone Percona MongoDB Exporter v0.51.0, compatibility mode enabled.
- MongoDB 8.3.3, FCV 8.0; Linux/Kubernetes.
- The same absolute-value calculation is present in v0.53.0 and in `main` at `49408b0b904b1b34b4e0924f85bf7ebe6b2e76fb`; the deterministic regression was run against that `main` revision.

**Additional context**

Related: #697 concerns unavailable members and zero timestamps. This case occurs with healthy members and nonzero timestamps, so it does not depend on that issue. A focused fix and regression tests will be linked in a PR.

tanzerm:
Fixes #1362.

When scraped on a secondary, `replSetGetStatus` can contain a local applied optime newer than the heartbeat-reported primary optime. The compatibility lag metric currently takes the absolute difference, so a secondary 10 seconds newer than that cached primary timestamp incorrectly reports 10 seconds of lag.

Calculate `max(0, primaryOptime - memberOptime)` instead. This preserves positive lag, the existing metric name and labels, and the guards that omit lag when no primary is found or either optime is unavailable. Remote-member data still depends on heartbeat freshness.

Add a deterministic test that exercises `replSetMetrics` and checks the emitted Prometheus metric for seven cases: newer secondary, caught-up secondary, lagging secondary, delayed secondary, no primary, missing primary optime, and missing secondary optime. It uses synthetic BSON data and needs no MongoDB connection. The newer-secondary case fails on unmodified `main` with 10 instead of 0; all seven pass with this fix.

Validation with Go 1.26.7:

- Regression plus the existing compatibility-conversion and metric unit tests passed, including with `-race`.
- `make format`, `make check-license`, and `git diff --check` passed.
- golangci-lint v2.13.1 reports zero issues introduced by this patch (`--new-from-rev=49408b0b904b1b34b4e0924f85bf7ebe6b2e76fb`). A full unfiltered lint run also reports existing findings outside the change.
- The Docker-based MongoDB integration suite was not run locally; upstream CI remains to be run.

Focused regression command:

```sh
go test -race ./exporter -run '^TestReplSetMetricsReplicationLag$' -count=1 -v
```

Related: #697 concerns unavailable members and zero timestamps; this fix addresses a different case involving healthy members with nonzero timestamps.

### DIFF
--- a/exporter/replset_lag_test.go
+++ b/exporter/replset_lag_test.go
@@ -1,0 +1,83 @@
+// mongodb_exporter
+// Copyright (C) 2026 Percona LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package exporter
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus/testutil"
+	"github.com/prometheus/common/promslog"
+	"github.com/stretchr/testify/require"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+)
+
+func TestReplSetMetricsReplicationLag(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name        string
+		primary     int64
+		secondary   int64
+		omitPrimary bool
+		wantLag     string
+	}{
+		{name: "secondary newer than primary heartbeat", primary: 1000, secondary: 1010, wantLag: "0"},
+		{name: "secondary caught up", primary: 1000, secondary: 1000, wantLag: "0"},
+		{name: "secondary behind primary", primary: 1017, secondary: 1000, wantLag: "17"},
+		{name: "delayed secondary", primary: 1060, secondary: 1000, wantLag: "60"},
+		{name: "no primary", secondary: 1000, omitPrimary: true},
+		{name: "primary optime unavailable", secondary: 1000},
+		{name: "secondary optime unavailable", primary: 1000},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			members := bson.A{bson.M{
+				"name":       "secondary:27017",
+				"stateStr":   "SECONDARY",
+				"health":     1,
+				"self":       true,
+				"optimeDate": primitive.DateTime(tc.secondary * 1000),
+			}}
+			if !tc.omitPrimary {
+				members = append(members, bson.M{
+					"name":       "primary:27017",
+					"stateStr":   "PRIMARY",
+					"health":     1,
+					"optimeDate": primitive.DateTime(tc.primary * 1000),
+				})
+			}
+
+			metrics := replSetMetrics(bson.M{"set": "rs0", "members": members}, promslog.New(&promslog.Config{}))
+			var expected string
+			if tc.wantLag != "" {
+				expected = fmt.Sprintf(`
+# HELP mongodb_mongod_replset_member_replication_lag The replication lag that this member has with the primary.
+# TYPE mongodb_mongod_replset_member_replication_lag gauge
+mongodb_mongod_replset_member_replication_lag{name="secondary:27017",self="1",set="rs0",state="SECONDARY"} %s
+`, tc.wantLag)
+			}
+
+			err := testutil.CollectAndCompare(staticCollector(metrics), strings.NewReader(expected), "mongodb_mongod_replset_member_replication_lag")
+			require.NoError(t, err)
+		})
+	}
+}
--- a/exporter/v1_compatibility.go
+++ b/exporter/v1_compatibility.go
@@ -1038,7 +1038,8 @@ func replSetMetrics(d bson.M, l *slog.Logger) []prometheus.Metric { //nolint:cyc
 				float64(m.ElectionTime.T), labels)
 		}
 		if t := m.OptimeDate.Time(); gotPrimary && t.Unix() != 0 && m.StateStr != "PRIMARY" {
-			val := math.Abs(float64(t.Unix() - primaryOpTime.Unix()))
+			// A member can be newer than the heartbeat-reported primary optime without being behind.
+			val := math.Max(0, float64(primaryOpTime.Unix()-t.Unix()))
 			createMetric("member_replication_lag",
 				"The replication lag that this member has with the primary.",
 				val, labels)


exporter/v1_compatibility.go
```diff
@@ -1038,7 +1038,8 @@ func replSetMetrics(d bson.M, l *slog.Logger) []prometheus.Metric { //nolint:cyc
 				float64(m.ElectionTime.T), labels)
 		}
 		if t := m.OptimeDate.Time(); gotPrimary && t.Unix() != 0 && m.StateStr != "PRIMARY" {
-			val := math.Abs(float64(t.Unix() - primaryOpTime.Unix()))
+			// A member can be newer than the heartbeat-reported primary optime without being behind.
+			val := math.Max(0, float64(primaryOpTime.Unix()-t.Unix()))
```
ademidoff, line 1042: The logic here is right. Optional style nit: `go.mod` is on `go 1.26.7`, so the builtin `max` works and lets the clamp stay in `int64` instead of round-tripping through `float64`:

```go
val := float64(max(0, primaryOpTime.Unix()-t.Unix()))
```

Note that this line is the only remaining `math.*` usage in the file, so switching would also mean dropping the `"math"` import (line 22) or the build breaks on an unused import. Entirely up to you — the current form passes lint as-is.

---
_Generated by [Claude Code](https://claude.ai/code)_

exporter/replset_lag_test.go
ademidoff, line 1: Nit on placement: `replSetMetrics` lives in `exporter/v1_compatibility.go` and its unit tests live in `exporter/v1_compatibility_test.go`, so that would be the natural home for this test. A new top-level `replset_lag_test.go` also sits confusingly close to the existing `exporter/secondary_lag_test.go`, which is the (currently skipped) integration test for the very same metric. Moving `TestReplSetMetricsReplicationLag` into `v1_compatibility_test.go` would keep the DB-free compatibility tests together; not a blocker either way.

---
_Generated by [Claude Code](https://claude.ai/code)_

exporter/replset_lag_test.go
```diff
@@ -0,0 +1,83 @@
...
+		omitPrimary bool
+		wantLag     string
+	}{
+		{name: "secondary newer than primary heartbeat", primary: 1000, secondary: 1010, wantLag: "0"},
+		{name: "secondary caught up", primary: 1000, secondary: 1000, wantLag: "0"},
+		{name: "secondary behind primary", primary: 1017, secondary: 1000, wantLag: "17"},
+		{name: "delayed secondary", primary: 1060, secondary: 1000, wantLag: "60"},
+		{name: "no primary", secondary: 1000, omitPrimary: true},
+		{name: "primary optime unavailable", secondary: 1000},
+		{name: "secondary optime unavailable", primary: 1000},
```
ademidoff, line 46: Two small readability suggestions for the table:

1. An empty `wantLag` doubles as "the metric must not be emitted at all", which is easy to miss — the last three cases rely on it. An explicit `wantAbsent bool` (or just a comment above those rows) would make the intent obvious to the next reader.
2. Every case here scrapes a `self: true` secondary. One extra row with a second, non-`self` secondary would lock in that heartbeat-sourced remote members still report their positive difference and that the `self="0"` label path is unaffected — that's the other half of the behaviour this change touches.

---
_Generated by [Claude Code](https://claude.ai/code)_

ademidoff (commented):
Thanks for the clear issue write-up and for pairing the fix with a deterministic regression test — that combination made this easy to verify. Leaving this as a comment rather than a verdict; @JiriCtvrtka and I will take the merge decision.

## What I verified locally (Go 1.26.7, `b3f572c`)

- `go test -race ./exporter -run '^TestReplSetMetricsReplicationLag$'` — all 7 subtests pass.
- Reverted **only** the production line back to `math.Abs(...)` and re-ran the new test: it fails exactly as described, `secondary newer than primary heartbeat` reporting `10` where `0` is expected. So the test is a genuine regression guard, not just coverage.
- `go build ./...`, `go vet ./exporter`, `gofmt -l exporter/`, and `make check-license` are all clean; the 2026 copyright header passes `.github/check-license.go`.
- The DB-free unit tests in the package (`TestWalkTo`, `TestMakeLockMetric`, `TestAddLocksMetrics`, `TestSumMetrics`, `TestCreateOldMetricFromNew`) pass with `-race`. The other tests in `exporter` need the Docker cluster and fail here for that reason only — unrelated to this change.
- Nothing else in the tree references `mongodb_mongod_replset_member_replication_lag` outside `secondary_lag_test.go` (skipped) and the new test, and `REFERENCE.md` does not document the metric, so no docs need updating.

## On the change itself

The sign fix is correct, and it's the right shape: `max(0, primary - member)` matches how MongoDB defines replication lag, keeps the metric non-negative (which was presumably the original intent behind `math.Abs`), and leaves the metric name, labels, and the "omit when no primary / either optime is zero" guards untouched. Dashboards and alert rules keep working; only the false-positive values change.

One consequence worth stating explicitly on the issue, since it's a behaviour trade-off rather than a pure bug fix: because the primary's `optimeDate` seen from a secondary is heartbeat-sourced (default `heartbeatIntervalMillis: 2000`), clamping means genuine lag smaller than that staleness window now reads `0` instead of a small positive number. That is clearly better than inverting the sign into a false alarm, but it does trade a false positive for a bounded false negative when scraped on a secondary. No change requested — just something a maintainer should be aware of when this lands.

Everything else is nits, inline below: the builtin `max` (plus the now-orphaned `"math"` import), test file placement, and two table-readability points.

## PR hygiene

- The branch is behind `main`. A rebase would be good before merge — it should also settle the red `codecov/project` check, which looks like a stale-base artifact rather than anything this patch caused (`codecov/patch` is green at 100% of new lines, and the Codecov comment itself notes the base report is 277 commits behind `main` with two fewer uploads). Build, Lint Check, and version-check are all green.
- Merges here squash with the PR title as the commit subject, and recent history uses conventional-commit prefixes (`fix:`, `ci:`, `chore(deps):`). Consider retitling to `fix: compatibility replication lag sign for stale primary optimes`. Not CI-enforced; a maintainer can also adjust it at merge time.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1364, opened 2026-09-08
Update mongo-go-driver to v2

bvalente:
Uses the latest mongo-go-driver at the date [v2.9.0](https://github.com/mongodb/mongo-go-driver/releases/tag/v2.9.0), and allows to use the official AWS SDK if using `MONGODB-AWS` from https://github.com/mongodb/mongo-go-driver/pull/2508


bvalente:
There are a lot of changes because of the linter

exporter/common.go
```diff
@@ -17,22 +17,38 @@ package exporter
...
-	"go.mongodb.org/mongo-driver/bson"
-	"go.mongodb.org/mongo-driver/bson/primitive"
-	"go.mongodb.org/mongo-driver/mongo"
-	"go.mongodb.org/mongo-driver/mongo/options"
+	"go.mongodb.org/mongo-driver/v2/bson"
+	"go.mongodb.org/mongo-driver/v2/mongo"
+	"go.mongodb.org/mongo-driver/v2/mongo/options"
 )
 
+const (
```
bvalente, line 30: the linter complained a lot about repeated keywords, so they were added as constants

exporter/exporter.go
```diff
@@ -425,6 +425,10 @@ func connect(ctx context.Context, opts *Opts) (*mongo.Client, error) {
 	if err != nil {
 		return nil, fmt.Errorf("invalid dsn: %w", err)
 	}
+	err = configureAWSCredentialsProvider(ctx, clientOpts)
```
bvalente, line 428: calling the function that loads and uses the aws sdk if MONGO-AWS is used

exporter/topology_info.go
```diff
@@ -120,9 +120,11 @@ func (t *topologyInfo) loadLabels(ctx context.Context) error {
...
 	}
 
-	cid, err := util.ClusterID(ctx, t.client)
-	if err != nil {
-		if nodeType != typeArbiter { // arbiters don't have a cluster ID
+	cid := ""
+	if nodeType != typeArbiter { // arbiters don't have a cluster ID
+		var err error
+		cid, err = util.ClusterID(ctx, t.client)
+		if err != nil {
```
bvalente, line 127: because of a different behavior in v2, we only ask for the cluster ID if we know it's not an arbiter

bvalente:
Hello @ademidoff @4nte I know this is a large patch, but I was hoping to contribute to this major version upgrade of the mongo-go-driver, which includes support for the native aws sdk library when using mongodb with aws iam

Besides all the changes to the code because of the library upgrade, I also had to do some lint cleanup as well as create more tests to pass all the GH checks.

Please let me know what you think of this contribution 😄

PR #1369, opened 2026-09-17
Build the GSSAPI test behind its tag, and fail loudly on a stopped test container

### LINKED ISSUE #1368: TestGSSAPIAuth: missing gssapi build tag, and a stopped container reported as an empty host
`TestGSSAPIAuth` is the one test in `exporter/` that cannot pass outside `make test`, and when its containers are missing it fails in a way that does not say so. Two separate causes.

### 1. No `gssapi` build tag

The mongo-driver gates GSSAPI behind `-tags gssapi`. Without it, `Auth` returns:

```
GSSAPI support not enabled during build (-tags gssapi)
```

`make test`, `make test-race` and `make test-cover` all pass the tag, so CI is fine. But no Go file in this repo carries that build constraint, so `TestGSSAPIAuth` is compiled and run by a plain `go test ./exporter/...` too, where it can only ever fail. It is also the only test requiring `libkrb5-dev` at build time, which CI installs explicitly and a contributor's machine may not have.

### 2. A stopped dependency is reported as an empty host

`IPForContainer` (`internal/tu/testutils.go:203`) reads `NetworkSettings` without checking the container's state. `docker inspect` succeeds for a container that exists but is not running, and the address field is empty, so the helper returns `("", nil)`.

The test then builds a URI around that empty host and fails further downstream — `unable to reach any KDC in realm PERCONATEST.COM, tried 0 KDCs` — which points at Kerberos rather than at the container that is not running.

### How this was hit

`psmdb-kerberos` would not start on Docker Desktop, whose VM kernel is `7.0.12-linuxkit`. MongoDB refuses to start on Linux kernels ≥ 6.19 ([SERVER-121912](https://jira.mongodb.org/browse/SERVER-121912)):

```
MongoDB cannot start: Linux kernel versions 6.19 and newer has a known incompatibility
with this version of MongoDB.
```

Worth noting this is not an architecture problem: `percona/percona-server-mongodb:8.0` publishes both `amd64` and `arm64`, and the image ran natively. Any machine with a new enough kernel sees the same thing, so the condition to detect is "the dependency is not serving", not "this platform is unsupported".

### Proposed fix

- Move `TestGSSAPIAuth` and its `generateKerberosConfigFile` helper into `exporter/gssapi_test.go` behind `//go:build gssapi`. CI and every `make test*` target already pass the tag, so no coverage is lost.
- Make `IPForContainer` fail when the container is not running, naming it.

Deliberately **not** proposing a skip when the container is absent: GSSAPI coverage would then disappear silently the day `psmdb-kerberos` breaks in CI. Failing loudly with an accurate message is the wanted behaviour.

ademidoff:
Fixes #1368.

`TestGSSAPIAuth` was the one test in `exporter/` that could not pass outside `make test`, and when its containers were missing it failed in a way that did not say so.

### Build the GSSAPI test only with `-tags gssapi`

The driver compiles GSSAPI in behind that tag and answers every handshake with `GSSAPI support not enabled during build` without it, so a plain `go test ./exporter/...` compiled the test and could only ever fail it. It also needs the Kerberos development headers at build time, which CI installs with `libkrb5-dev` and a contributor's machine may not have.

`TestGSSAPIAuth` and its `generateKerberosConfigFile` helper move to `exporter/gssapi_test.go` behind `//go:build gssapi`. **No coverage is lost** -- every `make test*` target and CI pass the tag:

```
$ go test -list TestGSSAPIAuth ./exporter/
ok      github.com/percona/mongodb_exporter/exporter

$ go test -tags gssapi -list TestGSSAPIAuth ./exporter/
TestGSSAPIAuth
ok      github.com/percona/mongodb_exporter/exporter
```

### Fail loudly when a test container is not running

`docker inspect` succeeds for a container that exists but has stopped, and the address and port fields are empty by then. Both `IPForContainer` and `PortForContainer` handed that empty string back with a **nil error** -- not by accident: `errors.Wrapf` returns nil when the error it wraps is nil, so every guard below the inspect reported success.

Before and after, against a container created and never started:

```
--- FAIL: TestContainerHelpersFailWhenContainerIsNotRunning
    --- FAIL: TestContainerHelpersFailWhenContainerIsNotRunning/IPForContainer
            Error:    An error is expected but got nil.
            Messages: a container that is not running was reported as fine
    --- FAIL: TestContainerHelpersFailWhenContainerIsNotRunning/PortForContainer
            Error:    An error is expected but got nil.
            Messages: a container that is not running was reported as fine
```

That is how a stopped `psmdb-kerberos` reached `TestGSSAPIAuth` as `unable to reach any KDC in realm PERCONATEST.COM, tried 0 KDCs` -- pointing at Kerberos rather than at the container that was down. Now:

```
--- FAIL: TestGSSAPIAuth (0.02s)
    gssapi_test.go:77:
        Error: Received unexpected error:
               container "psmdb-kerberos" is not running (state: exited, exit code: 1)
```

and it fails before connecting rather than after.

**No skip.** If a container of the test cluster is missing in CI that is worth failing over, and skipping would retire the GSSAPI coverage silently the day `psmdb-kerberos` breaks.

Two things came along because the guards needed rewriting anyway: the helpers move off `github.com/pkg/errors`, which `.golangci.yml` denies, and `GetImageNameForContainer` reported the literal `"mongo-1-1"` whatever container it was asked about.

### Verification

`make test` against the cluster: `internal/tu` green including the new test, and the only failures are the two this machine cannot run -- `TestGSSAPIAuth`, because MongoDB refuses to start on Docker Desktop's kernel ([SERVER-121912](https://jira.mongodb.org/browse/SERVER-121912)), which is what prompted all this, and `TestPBMCollector`, which needs `init-pbm.sh`. golangci-lint reports nothing on the added lines; `make check-license` and gofumpt/gci are clean.


JiriCtvrtka (approved):
Few things with low severity:
internal/tu/testutils.go:204 — errContainerMissing branch is unreachable (docker inspect on a missing container exits 1, so InspectContainer returns on the err != nil path); the make test-cluster hint never shows, and .Output() at internal/tu/testutils.go:172 swallows docker's "No such object" stderr, leaving only "exit status 1".
exporter/gssapi_test.go:80 — require.NoError(t, err) re-checks the stale err from tu.IPForContainer on line 76; it is a no-op that looks like a guard on krb5.conf creation.
exporter/gssapi_test.go:105 — assert.NoError (not require) on connect; on failure the test continues with a nil client and fails with a mongodb_up 0 metric diff instead of the auth error.
internal/tu/testutils.go:88 — GetImageNameForContainer was not converted to runningContainer, so a stopped container still returns its image name with a nil error — the same silent success the PR removes elsewhere.
exporter/gssapi_test.go:16 — the lint workflow runs golangci-lint without --build-tags gssapi, so the new file is never linted.

